PR #13, opened 2023-04-19
Cleanup karpenter values and refactor gpu provisioner in karpenter-config

chiragjn:
…

### DIFF
--- a/charts/tfy-karpenter-config/values.yaml
+++ b/charts/tfy-karpenter-config/values.yaml
@@ -18,9 +18,6 @@ karpenter:
       - key: kubernetes.io/arch
         operator: In
         values: ["amd64"]
-      - key: karpenter.k8s.aws/instance-family
-        operator: NotIn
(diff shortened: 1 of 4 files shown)


charts/tfy-karpenter-config/values.yaml
```diff
@@ -18,13 +18,16 @@ karpenter:
       - key: kubernetes.io/arch
         operator: In
         values: ["amd64"]
+      - key: karpenter.k8s.aws/instance-family
```
opsved, line 21: By default instance-family is '`c', 'm', 'r'.  [Default Instance type](https://karpenter.sh/v0.26.1/concepts/provisioners/#instance-types)

opsved (changes requested):

charts/tfy-karpenter-config/values.yaml
chiragjn: removed for now

opsved (approved):
LGTM

PR #25, opened 2023-05-19
Revert change

srihari-tf:


opsved (changes requested):
We might have to modify liveness probes and readiness probes or increase the memory request a little bit. I am submitting my changes on top of this

PR #26, opened 2023-05-23
New tfy agent

dunefro:


charts/tfy-agent/values.yaml
```diff
@@ -1,123 +1,296 @@
...
+  rbac:
+    enabled: true
 
-image:
-  repository: truefoundrycloud/tfy-agent
-  pullPolicy: IfNotPresent
-  # Overrides the image tag whose default is the chart appVersion.
-  tag: "a9af8a51d8b0373f4d0c469119ed88a488c5709f"
+# configuration for tfy-agent and tfy-agent-proxy
+config:
```
debajyoti-truefoundry, line 11: Normally these are done via the `.env.default` approach.

https://github.com/truefoundry/tfy-agent/blob/main/.env.default
https://github.com/truefoundry/helm-charts/blob/fd9a78ec2a7fc3b21f18cc610f1eff6d591b478c/charts/tfy-agent/values.yaml#L58

Going forward, do we need to also add a config field, if we introduce an env var?

charts/tfy-agent/values.yaml
```diff
@@ -1,123 +1,296 @@
...
-  liveness:
-    port: 3000
-    path: /
 
-nodeSelector: {}
-tolerations: {}
-affinity: {}
+  ## Number of replicas of tfyAgent
+  #
+  replicaCount: 1
```
debajyoti-truefoundry, line 202: We should not expose this option for `tfy-agent`/`tfy-agent-proxy` as of now. There should be strictly only 1 pod for these services.

debajyoti-truefoundry: Later if we want to implement a hot stand-by pod via leader election, we can expose this.

charts/tfy-agent/values.yaml
```diff
@@ -1,123 +1,296 @@
...
 
-envSecretName: tfy-agent-env-secret
-podAnnotations: {}
+  ## Image of tfyAgent
+  ##
+  image:
+    repository: truefoundrycloud/tfy-agent
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: "a9af8a51d8b0373f4d0c469119ed88a488c5709f"
```
debajyoti-truefoundry, line 114: Is this going to be automatically updated via our existing CI system? Or will this be manual?

https://github.com/truefoundry/tfy-agent/blob/a9af8a51d8b0373f4d0c469119ed88a488c5709f/.github/workflows/build-n-publish.yaml#L61

charts/tfy-agent/values.yaml
```diff
@@ -1,123 +1,296 @@
...
+        - path: /
+          pathType: ImplementationSpecific
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
+
+# Enabling virtual services for istio
+# disabled by default
+# required when tfy-agent is to be exposed outside
```
debajyoti-truefoundry, line 290: We will not need to expose Agent outside. In the current architecture, we should ensure that the Agent is not contactable from the outside cluster apart form our control-plane, which does not require a VS.

charts/tfy-agent/Chart.yaml
```diff
@@ -1,6 +1,24 @@
 apiVersion: v2
 name: tfy-agent
+description: A Helm chart for Kubernetes
```
srihari-tf, line 3: this is a placeholder description?

charts/tfy-agent/values.yaml
dunefro: We will introduce `extraEnvVars: []` feature for both applications in coming release.

debajyoti-truefoundry (approved):

charts/tfy-agent/values.yaml
```diff
@@ -1,123 +1,304 @@
...
-  tag: "a9af8a51d8b0373f4d0c469119ed88a488c5709f"
+# configuration for tfy-agent and tfy-agent-proxy
+config:
+  # token to connect to the contorl plane
+  clusterToken: ""
+
+  # name of the secret containing the cluster token
+  # the key must be token
+  # if this is set it will override config.clusterToken
+  clusterTokenSecret: ""
```
debajyoti-truefoundry, line 18: What should be the key name within the secret?

srihari-tf, line 18: if user provides a `clusterTokenSecret` the expectation might be that it already exists in the cluster. But current behaviour seems to be to create a new secret even when `clusterTokenSecret` is provided.

charts/tfy-agent/templates/_helpers.tpl
```diff
@@ -24,106 +24,75 @@ Expand the name of the chart.
...
-{{- if and $val (contains "${k8s-secret" ($val | toString)) }}
-- name: {{ $key }}
-  valueFrom:
-    secretKeyRef:
-      name: {{ $.Values.envSecretName }}
-      key: {{ $val | trimPrefix "${k8s-secret/" | trimSuffix "}" }}
+Create the name of the secret which will contain cluster token
+*/}}
+{{- define "tfy-agent.secretName" -}}
+{{- if .Values.config.clusterTokenSecret }}
```
srihari-tf, line 93: we can skip this condition now right?

PR #29, opened 2023-05-30
istio installation during cluster onboarding

kushagra-truefoundry:


scripts/bootstrap_cluster.sh
```diff
@@ -119,6 +119,29 @@ install_argocd_helm_chart() {
     --set controller.extraArgs[0]='--application-namespaces="*"'
 }
 
+install_istio() {
+    local tenant_name=$1
+    local cluster_token=$2
+    local ip_adress=$(kubectl cluster-info | head -n 1 | grep -oE '[^[:space:]]+' | sed -n '7p')
+
+    response=$(curl -X POST 'http://localhost:3000/v1/cluster-onboarding/configure-ingress' \
```
srihari-tf, line 127: replace localhost with control plane url?

scripts/bootstrap_cluster.sh
```diff
@@ -119,6 +119,29 @@ install_argocd_helm_chart() {
     --set controller.extraArgs[0]='--application-namespaces="*"'
 }
 
+install_istio() {
+    local tenant_name=$1
+    local cluster_token=$2
+    local ip_adress=$(kubectl cluster-info | head -n 1 | grep -oE '[^[:space:]]+' | sed -n '7p')
```
srihari-tf, line 125: can we verify this works for any cluster? @opsved

scripts/bootstrap_cluster.sh
srihari-tf, line 1: ask user for control plane url as the third argument - if it is not provided, set it to `https://$tenant_name.truefoundry.cloud`

scripts/bootstrap_cluster.sh
```diff
@@ -119,6 +119,29 @@ install_argocd_helm_chart() {
...
+            "clusterToken": "'$cluster_token'"
+        }' \
+        -w "%{http_code}" \
+        -o response.json \
+        --silent)
+    if [ "$response" == "201" ]; then
+        echo "Istio installed successfully"
+    else
+        echo "Failed to install istio response code: $response"
+        return 1
```
srihari-tf, line 141: exit here?

dunefro (commented):
This logic is wrong here [and]https://github.com/truefoundry/infra-charts/blob/fce7f7905cbf1391e398113800ad2b12a9689e17/scripts/bootstrap_cluster.sh#L47 should be OR `||`

scripts/bootstrap_cluster.sh
sri-rad: curl is not to check - it is to install istio and create DNS record

PR #39, opened 2023-06-06
added logic to install istio with values file

kushagra-truefoundry:


scripts/bootstrap_cluster.sh
```diff
@@ -119,10 +133,22 @@ install_argocd_helm_chart() {
     --set controller.extraArgs[0]='--application-namespaces="*"'
 }
 
+save_istio_values() {
```
srihari-tf, line 136: we are only doing it for tfy-istio-ingress? what about other charts?

scripts/bootstrap_cluster.sh
```diff
@@ -119,10 +133,22 @@ install_argocd_helm_chart() {
...
+    local cluster_type=$1
+    response=$(curl "https://catalogue.truefoundry.com/$cluster_type/templates/istio/tfy-istio-ingress.yaml")
+    echo "$response" > application.yaml
+    yq '.spec.source.helm.values' application.yaml > values.yaml
```
srihari-tf, line 140: can we name it more accurately - so that it's not generic `values.yaml`

scripts/bootstrap_cluster.sh
```diff
@@ -151,8 +177,9 @@ installation_guide() {
         # Istio CRDs are already installed, skip the entire Istio installation
         print_yellow "Skipping istio charts installation."
     else
+        save_istio_values "$cluster_type"
         helm repo add istio https://istio-release.storage.googleapis.com/charts
-        install_helm_chart "istio" "base" "istio-system" "1.15.3"
+        install_helm_chart_with_values "istio" "base" "istio-system" "1.15.3" "values.yaml"
```
srihari-tf, line 182: where are we installing `tfy-istio-ingress` - is this values correct?

scripts/bootstrap_cluster.sh
kushagra-truefoundry: removed these

srihari-tf (dismissed):

scripts/bootstrap_cluster.sh
```diff
@@ -124,7 +129,7 @@ install_helm_chart_with_values() {
 
 
 install_argocd_helm_chart() {
-    helm install argocd argo/argo-cd --version 5.16.13 \
+    helm install argod argo/argo-cd --version 5.16.13 \
```
kushagra-truefoundry, line 132: we are directly installing argo/argo-cd, without using helm repo add .. , what if initially the machine do not have that chart

scripts/bootstrap_cluster.sh
```diff
@@ -177,32 +207,26 @@ installation_guide() {
...
-    fi
-    
-    if check_istio_crds_installed; then
-        # Istio CRDs are already installed, skip the entire Istio installation
-        print_yellow "Skipping istio charts installation."
-    else
-        install_istio_dependencies "$cluster_type"
+    helm repo add argo https://argoproj.github.io/argo-helm
+    install_argocd_helm_chart
+    install_argo_charts "$cluster_type"
```
kushagra-truefoundry, line 218: I think a tab space will be good here.

PR #66, opened 2023-07-13
Bug fix for default provisioner providerref

dunefro:


charts/tfy-karpenter-config/values.yaml
```diff
@@ -24,7 +24,7 @@ karpenter:
       - key: karpenter.k8s.aws/instance-size
         operator: NotIn
         values: ["nano", "micro", "metal"]
-    providerRefName: default
+  defaultProvisionerProviderRefName: default
```
chiragjn, line 27: probably we can just do
```
karpenter:
  defaultProvisionerSpec:
    providerRefName:
      name: default
```

and then delete line 8 and 9 in `charts/tfy-karpenter-config/templates/karpenter-default-provisioner.yaml`

chiragjn (approved):

chiragjn (approved):

PR #71, opened 2023-07-25
Fix notebook controller chart

akashg3627:


charts/tfy-notebook-controller/templates/notebook-controller-standalone.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     app: notebook-controller
     control-plane: controller-manager
     kustomize.component: notebook-controller
-  name: nb-ctl-system
+  name: {{ .Release.Namespace }}
```
dunefro, line 8: We should remove the creation of namespace from the chart

dunefro (changes requested):

charts/tfy-notebook-controller/templates/notebook-controller-standalone.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
     app: notebook-controller
```
debajyoti-truefoundry, line 5: Do we need this whole block?

debajyoti-truefoundry: In the flow, namespace will be already created, right?

debajyoti-truefoundry: Is this required to add the labels?

PR #81, opened 2023-08-07
fix: namespace issue in tfy-notebook-controller-chart

akashg3627:


nikp1172 (approved):

dunefro (commented):
If you are making this change, let's make all these separate files and derive the values from `values.yaml` file

dunefro (approved):
This looks good on the whole level just the fact some file are named directly `role.yaml`, If possible please make it `notebook-controller-role.yaml`,
Similarly for `leader-election-rolebinding.yaml` and `notebook-controller-rolebinding.yaml`

charts/tfy-notebook-controller/values.yaml
```diff
@@ -1 +1,15 @@
-
+labels:
+  app: notebook-controller
+  kustomize.component: notebook-controller
+namespace: '{{ .Release.Namespace }}'
```
dunefro, line 4: this is not needeed

charts/tfy-notebook-controller/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+              path: /readyz
+              port: 8081
+            initialDelaySeconds: 5
+            periodSeconds: 10
+      serviceAccountName: notebook-controller-service-account
+      tolerations:
+        - key: CriticalAddonsOnly
+          value: "true"
+          effect: NoSchedule
+          operator: Equal
```
dunefro, line 78: Let's move these tolerations out if possible. We will pass them separately.

charts/tfy-notebook-controller/templates/cluster-role-binding.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{- end }}
+  name: notebook-controller-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: notebook-controller-role
+subjects:
+  - kind: ServiceAccount
+    name: notebook-controller-service-account
+    namespace: {{ tpl .Values.namespace . }}
```
dunefro, line 16: You can directly pass the release.namespace here

PR #91, opened 2023-08-22
Pass device specs in device plugin on Azure

chiragjn:
See: https://github.com/Azure/AKS/issues/3680#issuecomment-1585298045
This fixes an issue when `SystemdCgroup` is enabled for containerd - which is the default on AKS 1.25+ running Ubuntu 22.04
In addition to this, it is necessary to run at least  VM image `202306.07.0` because that has udev rules to create symlinks in `/dev/char`

### DIFF
--- a/charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
+++ b/charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
@@ -56,6 +56,11 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
+          env:
+            - name: FAIL_ON_INIT_ERROR
+              value: "false"
+            - name: PASS_DEVICE_SPECS
+              value: "true"
           volumeMounts:
             - name: device-plugin
               mountPath: /var/lib/kubelet/device-plugins
--- a/charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
+++ b/charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
@@ -50,17 +50,15 @@ spec:
           value: spot
           effect: NoSchedule
       containers:
-        - image: mcr.microsoft.com/oss/nvidia/k8s-device-plugin:1.11
+        - image: nvcr.io/nvidia/k8s-device-plugin:v0.13.0-rc.1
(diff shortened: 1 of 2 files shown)


charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
```diff
@@ -56,11 +56,16 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
+          env:
```
chiragjn, line 59: I am not sure if the `mcr.microsoft.com/oss/nvidia/k8s-device-plugin:1.11` supports it, I'll test, if it does not work then I'll switch to `nvcr.io/nvidia/k8s-device-plugin:v0.14.1`

charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
```diff
@@ -50,17 +50,20 @@ spec:
           value: spot
           effect: NoSchedule
       containers:
-        - image: mcr.microsoft.com/oss/nvidia/k8s-device-plugin:1.11
+        - image: nvcr.io/nvidia/k8s-device-plugin:v0.14.1
```
chiragjn, line 53: yes, this versioning is correct, it is not a downgrade

charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
```diff
@@ -50,17 +50,20 @@ spec:
...
           name: nvidia-device-plugin-ctr
+          env:
+            - name: FAIL_ON_INIT_ERROR
```
shubhamrai1993, line 56: Is there a reason for this?

shubhamrai1993 (approved):

charts/tfy-gpu-operator/templates/aks-nvidia-device-plugin.yaml
chiragjn: I followed the recommendation in the linked comment in description. Actually now you mention it, let me remove it,
Fail on init false is required to run the daemonset on nodes that don't have gpus which we have avoided via affinity anyway

Thanks for bringing this up

PR #98, opened 2023-08-30
Control plane provisioner added

dunefro:


charts/tfy-karpenter-config/values.yaml
```diff
@@ -93,6 +94,61 @@ karpenter:
...
+    zones: []
+
+    instanceFamilies:
+      ## @param karpenter.controlPlaneProvisioner.instanceFamilies.allowed Allowed instance families for control plane workloads
+      ##    
+      allowed: []
+
+      ## @param karpenter.controlPlaneProvisioner.instanceFamilies.notAllowed Not allowed instance families for control plane workloads
+      ##      
+      notAllowed: ["t3", "t2", "t3a", "p2", "p3", "p4d", "p4de", "g4dn", "g5", "g4ad", "inf1", "inf2", "trn1", "trn1n"]
```
shubhamrai1993, line 129: Is there a reason for not allowing the burstable ones?

shubhamrai1993 (approved):

charts/tfy-karpenter-config/values.yaml
dunefro: performance issues with burstable ones

PR #115, opened 2023-09-21
tfy-gpu-operator - Isolate GPUs correctly among pods running on the same node

chiragjn:
Changelog:

`tfy-karpenter-config`
---

### Major Changes:
- Introduces a new AWSNodeTemplate for gpu nodes with custom user data script. This script writes a custom containerd config to `/etc/eks/containerd/containerd-config.toml` with `plugins.cri.containerd.default_runtime_name` added and `plugins.cri.containerd.default_runtime.*` removed.

> How does this work?
> 1. EKS bootstrap.sh will copy it to `/etc/containerd/config.toml` - https://github.com/awslabs/amazon-eks-ami/blob/2b5e6eea3695747780434c5d47750fc27a7685c1/files/bootstrap.sh#L578-L585
> 2. `nvidia-container-toolkit` will install itself and update the nvidia Runtime binary in `/etc/containerd/config.toml`. This is important because even though EKS GPU ami comes with its own nvidia-container-toolkit installed we can't customize its behavior to use volume mounts strategy
> 3. Because we removed `plugins.cri.containerd.default_runtime.*`, now containerd will correctly use `nvidia` runtime with the intended container toolkit binary from step 2 - https://github.com/containerd/containerd/blob/1fbd70374134b891f97ce19c70b6e50c7b9f4e0d/pkg/cri/config/config.go#L413-L423

This is a temporary solution until the EKS team updates the AMIs (https://github.com/awslabs/amazon-eks-ami/issues/1329), we'll have to keep an eye on when a new AMI with config version 2 is released. We could have used version 2 config, but at this point, we want to keep the changes minimal. We are only doing this when containerd version is >=1.5.0 and <2.0.0, so we should be relatively safe

### Minor Changes:
- Add `p5` instance types in affinities

`tfy-gpu-operator`
---
### Major Changes:
- Device plugin will now list devices using index instead of uuid
- Device plugin will now pass devices to the container toolkit as volume mounts 
- Container Toolkit will now accept devices only as volume mounts. Env variable `NVIDIA_VISIBLE_DEVICES=all` is ignored.
- Pods that do not request any GPUs (hence do not get any volume mounts) can still request GPUs using `NVIDIA_VISIBLE_DEVICES=all` if they run in privileged mode

### Minor Changes:
- Put GC pod on non gpu nodes
- Reduces memory request on GC pod
- Add `p5` instance types in affinities

---

Tests
---
| g4dn.12xlarge - 4 GPUS |           |                        |                  |
|------------------------|-----------|------------------------|------------------|
| Privileged?            | GPU Count | NVIDIA_VISIBLE_DEVICES | Per pod Result?  |
| No                     | 2         | all                    | 2 GPUs assigned  |
| No                     | 2         | none                   | 2 GPUs assigned  |
| No                     | 2         | void                   | 2 GPUs assigned  |
| No                     | 2         | 0,2,3                  | 2 GPUs assigned  |
| No                     | 0         | all                    | No GPUs assigned |
| No                     | 0         | none                   | No GPUs assigned |
| No                     | 0         | void                   | No GPUs assigned |
| No                     | 0         | 0,2,3                  | No GPUs assigned |
| Yes                    | 2         | all                    | All 4 assigned   |
| Yes                    | 2         | none                   | All 4 assigned   |
| Yes                    | 2         | void                   | All 4 assigned   |
| Yes                    | 2         | 0,2,3                  | All 4 assigned   |
| Yes                    | 0         | all                    | All 4 assigned   |
| Yes                    | 0         | none                   | All 4 assigned   |
| Yes                    | 0         | void                   | No GPUs assigned |
| Yes                    | 0         | 0,2,3                  | All 4 assigned   |
---
⚠️ GPU Isolation will continue to be broken for nodes provisioned using EKS nodegroups. For that, we need to adjust the launch template with a similar user data script - https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html


charts/tfy-gpu-operator/values.yaml
```diff
@@ -117,7 +117,14 @@ aws-eks-gpu-operator:
       resources:
         requests:
           cpu: 10m
-          memory: 300Mi
+          memory: 100Mi
+      tolerations:
+        ## @skip aws-eks-gpu-operator.node-feature-discovery.gc.tolerations[0]
+        - key: nvidia.com/gpu
+          operator: Exists
+          effect: PreferNoSchedule
```
chiragjn, line 125: I am a little concerned if GC gets put on a GPU node and it might cause Karpenter to not drop the GPU node when consolidation is turned off.
Should I force gc to be on non-GPU nodes via a NotIn instance size affinity?

debajyoti-truefoundry: > Should I force gc to be on non-GPU nodes via a NotIn instance size affinity?

I think this is a safer option without loosing much.

charts/tfy-karpenter-config/templates/karpenter-gpu-default-awsnodetemplate.yaml
```diff
@@ -0,0 +1,78 @@
...
+    karpenter.sh/discovery: {{ .Values.cluster.name }}
+  {{- range $key, $val := .Values.karpenter.defaultNodeTemplate.extraTags }}
+    {{ $key }}: {{ $val }} 
+  {{- end }}
+  blockDeviceMappings:
+    - deviceName: /dev/xvda
+      ebs:
+        volumeSize: 100Gi
+        volumeType: gp3
+        deleteOnTermination: true
```
chiragjn, line 23: We need to expose these in Values in some sane way
Not in this PR though

debajyoti-truefoundry (approved):

PR #127, opened 2023-11-01
Replace tfy-gpu-operator manifests with helm charts, bump gpu-operator, dcgm exporter

chiragjn:
AWS EKS
---
- Bumps `gpu-operator`
  - Bumps container toolkit
  - Bumps device plugin
  - Bumps dcgm exporter
  - Deletes NFD GC template because it is installed by the `gpu-operator` chart itself

GCP GKE
---
- Bumps dcgm exporter

Azure AKS
---
- Replaces device plugin template with helm chart
- Bumps dcgm exporter

TODO:
- [ ] Testing
- [ ] Update help strings in values.yaml


charts/tfy-gpu-operator/values.yaml
```diff
@@ -339,6 +332,45 @@ gcp-gke-standard-dcgm-exporter:
...
+##
+azure-aks-nvidia-device-plugin:
+  namespaceOverride: tfy-gpu-operator
+  updateStrategy:
+    type: OnDelete
+  podAnnotations:
+    scheduler.alpha.kubernetes.io/critical-pod: ""
+  compatWithCPUManager: true
+  deviceListStrategy: volume-mounts
+  deviceIDStrategy: index
```
chiragjn, line 345: These are new additions which were not enabled before, will check if they make a difference

charts/tfy-gpu-operator/Chart.yaml
```diff
@@ -1,22 +1,27 @@
...
-    version: v23.3.1
+    version: v23.9.0
     alias: aws-eks-gpu-operator
   - condition: clusterType.gcpGkeStandard
     name: dcgm-exporter
     repository: https://nvidia.github.io/dcgm-exporter/helm-charts
-    version: 3.1.2
+    version: 3.1.7
     alias: gcp-gke-standard-dcgm-exporter
+  - condition: clusterType.azureAks
```
shubhamrai1993, line 18: Should we use gpu-operator - https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/23.9.0/microsoft-aks.html . That feels like the recommended option

chiragjn: Yeah, not a bad idea, will try it out

chiragjn:
We'll rework this later

PR #132, opened 2023-11-17
feat: add inferentia provisioner

debajyoti-truefoundry:


charts/tfy-karpenter-config/values.yaml
```diff
@@ -222,3 +222,40 @@ karpenter:
...
+    ## @param karpenter.inferentiaProvisionerSpec.instanceFamilies Instance families to launch instances for inferentia provisioner
+    ## 
+    instanceFamilies:
+    - inf1
+    - inf2
+    ## @param karpenter.inferentiaProvisionerSpec.instanceSizes.notAllowed Instance Sizes that are not allowed to launch instances for inferentia provisioner
+    ##
+    instanceSizes:
+      notAllowed:
+      - 48xlarge
```
chiragjn, line 258: interesting, any reason why?

debajyoti-truefoundry: Most of our customers will not need it. We had similar feedback where some customers panicked when Karpenter brought up big machines.

debajyoti-truefoundry: This is the most compute-heavy and expensive size.

charts/tfy-karpenter-config/values.yaml
```diff
@@ -222,3 +222,40 @@ karpenter:
...
+    ## @param karpenter.inferentiaProvisionerSpec.providerRefName Name of AWS node template to be used for inferentia provisioner
+    ##
+    providerRefName: default
```
dunefro, line 261: I am not sure if the `default` awsnodetemplate will be right for this. We can make sure if disk requirements is fine.

charts/tfy-karpenter-config/templates/karpenter-inferentia-provisioner.yaml
```diff
@@ -0,0 +1,49 @@
...
+      {{- end }}
+    - key: karpenter.k8s.aws/instance-size
+      operator: NotIn
+      values:
+      {{- range .Values.karpenter.inferentiaProvisionerSpec.instanceSizes.notAllowed  }}
+        - {{ . | quote }}
+      {{- end }}
+  taints:
+    - key: "aws.amazon.com/neuron"
+      effect: "NoSchedule"
```
dunefro, line 46: Are we making sure these taints are handled by the tolerations from the platform.

charts/tfy-karpenter-config/values.yaml
```diff
@@ -222,3 +222,40 @@ karpenter:
...
+  ## provisioner for inferentia workloads
+  inferentiaProvisionerSpec:
+    ## @param karpenter.inferentiaProvisionerSpec.enabled Enable inferentia provisioner for inferentia nodes
+    ##
+    enabled: false
+
+    ## @param karpenter.inferentiaProvisionerSpec.consolidation.enabled Enable consolidation for inferentia provisioner
+    ##
+    consolidation:
+      enabled: false
```
dunefro, line 235: Any reason to keep this `false` by default ?

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: Fine for now. I have tested xlarge, and it works fine. I will update this config if I see any issues in further testing.

charts/tfy-karpenter-config/templates/karpenter-inferentia-provisioner.yaml
debajyoti-truefoundry: We will ensure it.

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: Yes. It is not tested yet. I could not bring up larger machines having multiple accelerators due to quota unavailability.

charts/tfy-karpenter-config/templates/karpenter-inferentia-provisioner.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: inferentia-provisioner
+  namespace: karpenter
+spec:
+  weight: 10
+  {{- if .Values.karpenter.inferentiaProvisionerSpec.consolidation.enabled }}
+  consolidation:
+    enabled: true
+  {{- else }}
+  ttlSecondsAfterEmpty: {{ .Values.karpenter.inferentiaProvisionerSpec.ttlSecondsAfterEmpty }}
+  {{- end }}
```
chiragjn, line 14: Not directly relevant to this PR, 
I have seen karpenter complains that in case of gpu provisioner it got both `consolidation` and `ttlSecondsAfterEmpty` keys and spec is invalid
Now considering this if else it should be impossible
so my theory is argocd has some weird bug where this if condition is not evaluated correctly

dunefro (approved):

charts/tfy-karpenter-config/values.yaml
```diff
@@ -222,3 +222,40 @@ karpenter:
...
+
+    ## @param karpenter.inferentiaProvisionerSpec.ttlSecondsAfterEmpty Seconds after which node should be deleted once it is empty. Either one of consolidation or ttlSecondsAfterEmpty should be used
+    ## 
+    ttlSecondsAfterEmpty: 300
+
+    ## @param karpenter.inferentiaProvisionerSpec.capacityTypes Capacity types for inferentia provisioner
+    ##    
+    capacityTypes:
+    - spot
```
chiragjn, line 244: We really need to have a linter with rule on how to indent yaml lists 🥲

charts/tfy-karpenter-config/templates/karpenter-inferentia-provisioner.yaml
dunefro: its not actually or I am not pretty sure, but from argo side it first tries to check the new change and then remove the old one

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: I suggest we add a pre-commit hook.

PR #149, opened 2023-11-24
fix: reduce the time pods are waiting in unschedulable queue

debajyoti-truefoundry:


charts/tfy-inferentia-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: tfy-inferentia-operator
-version: 0.2.0
+version: 0.2.1-rc1
```
chiragjn, line 3:
```suggestion
version: 0.2.1-rc.1
```

chiragjn (approved):

PR #159, opened 2023-12-08
ops: enable soci for non-gpu nodes

debajyoti-truefoundry:


charts/tfy-karpenter-config/values.yaml
```diff
@@ -157,7 +157,108 @@ karpenter:
...
+      [plugins."io.containerd.grpc.v1.cri".containerd]
+      default_runtime_name = "runc"
+      discard_unpacked_layers = true
+      ## ENABLE SOCI SNAPSHOTTER
+      snapshotter = "soci"
+      disable_snapshot_annotations = false
+      ##
+
+      [plugins."io.containerd.grpc.v1.cri"]
+      sandbox_image = "602401143452.dkr.ecr.eu-west-1.amazonaws.com/eks/pause:3.5"
```
chiragjn, line 231: This should not be hardcoded here!, instead use the pause template variable here and let bootstrap script replace this according to the region

chiragjn, line 231:
```suggestion
      sandbox_image = "SANDBOX_IMAGE"
```

charts/tfy-karpenter-config/values.yaml
```diff
@@ -157,7 +157,108 @@ karpenter:
...
+
+      ## SOCI PLUGIN
+      [proxy_plugins]
+      [proxy_plugins.soci]
+      type = "snapshot"
+      address = "/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock"
+      ##
+      EOF
+
+      systemctl restart containerd
```
chiragjn, line 254: Is this required? I think containerd is not started at this point

debajyoti-truefoundry: Yes, weird. For whatever reason, it was not working without this restart. I will check again.

charts/tfy-karpenter-config/values.yaml
```diff
@@ -157,7 +157,108 @@ karpenter:
...
+      Restart=always
+      RestartSec=5
+
+      [Install]
+      WantedBy=multi-user.target
+      EOF
+
+        systemctl daemon-reload
+        systemctl enable --now soci-snapshotter
+        systemctl status soci-snapshotter
```
chiragjn, line 211: I think when bootstrap script starts containerd it should also start soci snapshotter right? if we configure that containerd needs soci snapshotter as a dependency?

debajyoti-truefoundry: I do not want to patch containerd's service definition.

chiragjn (approved):

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: I do not know. Maybe it will work. The goal here is to see whether the soci-snapshooter is up without any issues.
https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#Exit%20status

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: Yeah, my bad. I found the logic on https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh

charts/tfy-karpenter-config/values.yaml
debajyoti-truefoundry: > The goal here is to see whether the soci-snapshooter is up without any issues.

If there are any issues, I will not even patch containerd.

charts/tfy-karpenter-config/values.yaml
```diff
@@ -157,7 +157,107 @@ karpenter:
...
+          ! version_lte "$2" "$1"
+      }
+
+      CONTAINERD_VERSION_OUTPUT="$(containerd --version)"
+      IFS=" " read -a  CONTAINERD_VERSION_STRING <<< "${CONTAINERD_VERSION_OUTPUT}"
+      CONTAINERD_VERSION=${CONTAINERD_VERSION_STRING[2]}
+      echo containerd version "${CONTAINERD_VERSION}"
+
+      setup_soci() {( set -ex
+        yum install fuse -y
```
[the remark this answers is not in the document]
chiragjn: Nevermind I think yum is fine
AL2023 points yum to dnf
https://docs.aws.amazon.com/linux/al2023/ug/package-management.html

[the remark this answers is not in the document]
debajyoti-truefoundry: Do you have documentation for reference? This already pulls packages from indexes maintained by AWS.

shubhamrai1993 (approved):

charts/tfy-karpenter-config/values.yaml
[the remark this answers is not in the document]
debajyoti-truefoundry: > AL2 has yum and AL2023 points yum to dnf

Yeah, the one that we are using right now does not even have the `dnf` binary.

PR #174, opened 2023-12-23
tfy-gpu-operator: Update tolerations for eks nfd gc and civo device plugin

chiragjn:


charts/tfy-gpu-operator/values.yaml
```diff
@@ -523,91 +525,9 @@ civo-talos-nvidia-device-plugin:
   ## @skip civo-talos-nvidia-device-plugin.tolerations
   ##
   tolerations:
-    - operator: Exists
-
-
-civo-talos-dcgm-exporter:
```
chiragjn, line 529: This is okay to remove because we have removed dcgm component in last release. We will bring it back using the gpu operator route later

dunefro (approved):

PR #176, opened 2024-01-03
EKS GPUs - Disable GSP from userData script

chiragjn:
This is a hacky but working (so far) solution to https://github.com/NVIDIA/open-gpu-kernel-modules/issues/446 and https://github.com/awslabs/amazon-eks-ami/issues/1523

### DIFF
--- a/charts/tfy-karpenter-config/values.yaml
+++ b/charts/tfy-karpenter-config/values.yaml
@@ -227,6 +227,17 @@ karpenter:
           echo "CONTAINERD_VERSION is empty or not within the specified range."
       fi
 
+      rmmod nvidia_drm
+      rmmod nvidia_modeset
+      rmmod nvidia_uvm
+      rmmod nvidia
+      echo "Writing NVreg_EnableGpuFirmware=0 to /etc/modprobe.d/nvidia.conf"
+      echo "options nvidia NVreg_EnableGpuFirmware=0" | tee --append /etc/modprobe.d/nvidia.conf
+      echo "Writing NVreg_EnableGpuFirmware=0 to /etc/modprobe.d/nvidia-gsp.conf"
+      echo "options nvidia NVreg_EnableGpuFirmware=0" | tee --append /etc/modprobe.d/nvidia-gsp.conf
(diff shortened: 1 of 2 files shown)


charts/tfy-karpenter-config/values.yaml
```diff
@@ -227,6 +227,17 @@ karpenter:
           echo "CONTAINERD_VERSION is empty or not within the specified range."
       fi
 
+      rmmod nvidia_drm
+      rmmod nvidia_modeset
+      rmmod nvidia_uvm
+      rmmod nvidia
```
chiragjn, line 233: This is on a best-effort basis which is why this section is at the end of the script. `rmmod` for a module fails if some processes are actively using the module or any of its dependents have not been unloaded before.

charts/tfy-karpenter-config/values.yaml
```diff
@@ -227,6 +227,17 @@ karpenter:
...
+      echo "Writing NVreg_EnableGpuFirmware=0 to /etc/modprobe.d/nvidia.conf"
+      echo "options nvidia NVreg_EnableGpuFirmware=0" | tee --append /etc/modprobe.d/nvidia.conf
+      echo "Writing NVreg_EnableGpuFirmware=0 to /etc/modprobe.d/nvidia-gsp.conf"
+      echo "options nvidia NVreg_EnableGpuFirmware=0" | tee --append /etc/modprobe.d/nvidia-gsp.conf
+      echo "Running dracut"
+      dracut -f
```
chiragjn, line 239: This will have no effect until reboot if the module was not unloaded beforehand

debajyoti-truefoundry: Is EKS fine with this?
How much latency are we adding here?

chiragjn: > Is EKS fine with this?
Does not cause any problems as such but the problem is this works now but if some config changes under our feet then it would stop working. 


> How much latency are we adding here?
Maybe 1-2s, I can measure it out

PR #180, opened 2024-01-09
Use standalone dcgm instead of embedded nvhostengine

chiragjn:
TODO:
- [x] Docs


debajyoti-truefoundry (approved):

charts/tfy-gpu-operator/values.yaml
```diff
@@ -20,6 +20,12 @@ clusterType:
...
 aws-eks-gpu-operator:
+  nfd:
+    enabled: true
+  
+  gfd:
+    enabled: true
+
   ## Driver configuration
   ## @param aws-eks-gpu-operator.driver.enabled Enabled/Disable driver installation.
   driver:
```
chiragjn, line 31: This disabling needs testing actually, might cause the toolkit pods to be in pending

charts/tfy-gpu-operator/templates/civo-talos-fake-toolkit-ready.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - name: main
+          image: alpine:3.19
+          command:
+            - sh
+            - '-c'
+          args:
+            - |-
+              set -ex;
+              touch /run/nvidia/validations/host-driver-ready;
+              touch /run/nvidia/validations/toolkit-ready;
```
chiragjn, line 31: This is needed till gpu-operator adds support to disable validations via init containers

PR #198, opened 2024-02-04
add tolerations to values

srihari-tf:


charts/tfy-notebook-controller/values.yaml
```diff
@@ -91,31 +91,8 @@ oauth:
 sdsServer:
   ## @param sdsServer.image image for sds-server
   image: "public.ecr.aws/truefoundrycloud/sds-server:cb6f4be9a5e342ad7a3ba4eb72bcc158501500d8"
+  ## @param sdsServer.affinity [object] Node affinity for sds-server
   affinity:
```
debajyoti-truefoundry, line 95: Don't you have to remove this field?

charts/tfy-notebook-controller/templates/oauth-secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.oauth.enabled -}}
```
debajyoti-truefoundry, line 1: What happens if the `.Values.oauth` field is absent? Usually, do we need to handle that edge case?

charts/tfy-notebook-controller/values.yaml
srihari-tf: i have removed the value but good to have since we might need it for aws

debajyoti-truefoundry: If the template already supports the key, why must we have it incomplete in the values file?

```suggestion
  affinity: {}
```

charts/tfy-notebook-controller/values.yaml
```diff
@@ -145,6 +122,16 @@ sdsServer:
       ephemeral-storage: 10M
       ## @param sdsServer.resources.requests.memory The minimum memory resources requested.
       memory: 30M
+  ## @param sdsServer.tolerations [array] Spot tolerations for the notebook controller deployment.
+  tolerations:
```
debajyoti-truefoundry, line 126: Can we please also add the number of replicas in the values file?

debajyoti-truefoundry (approved):

PR #200, opened 2024-02-06
enable fractional gpu for aws

nikp1172:


charts/tfy-gpu-operator/templates/fractional-gpu-timeslicing.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.clusterType.azureAks }}
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: time-slicing-config
```
shubhamrai1993, line 6: Lets not hardcode but make it a variable and pick from values. Same for the number of replicas at line #15

charts/tfy-gpu-operator/templates/fractional-gpu-timeslicing.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.clusterType.azureAks }}
```
chiragjn, line 1: This seems incorrect?

charts/tfy-gpu-operator/values.yaml
```diff
@@ -139,6 +142,10 @@ aws-eks-gpu-operator:
         - key: nvidia.com/gpu
           effect: NoSchedule
           operator: Exists
+        - key: nvidia.com/fractional-gpu
```
chiragjn, line 145: Maybe we should probably call this something like `nvidia.com/time-sliced-gpu` to be specific?

charts/tfy-gpu-operator/templates/fractional-gpu-timeslicing.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: time-slicing-config
+  namespace: {{ .Release.Namespace }}
+data:
+  fractional-10: |-
+    version: v1
```
chiragjn, line 10: Does this apply globally or only to certain nodes?

PR #207, opened 2024-02-19
Added probes and deployment strategy to sds server

shubhamrai1993:


charts/tfy-notebook-controller/values.yaml
```diff
@@ -76,7 +76,7 @@ notebookBaseDomainUrl: <to_be_provided>
 ## @param oauth.hmac HMAC key used for encoding/decoding tokens in OAuth2. This key is crucial for maintaining token integrity and security.
 ##
 oauth:
-  enabled: false
+  enabled: true
```
debajyoti-truefoundry, line 79:
```suggestion
  enabled: false
```

Let us not enable this by default as of now. @srihari-tf

sri-rad (approved):

PR #213, opened 2024-02-22
expose extra env in Azure Nvidia Device Plugin

nikp1172:


charts/tfy-gpu-operator/values.yaml
```diff
@@ -390,6 +390,9 @@ gcp-gke-standard-dcgm-exporter:
       mountPath: /etc/nvidia
 
 
+azure-aks-device-plugin:
```
chiragjn, line 393:
```suggestion
azure-aks-nvidia-device-plugin:
```

chiragjn:
Superseded by #214

PR #215, opened 2024-02-26
add spot toleration in azure nfd master

nikp1172:


charts/tfy-gpu-operator/values.yaml
```diff
@@ -455,6 +455,29 @@ azure-aks-gpu-operator:
...
+          operator: Exists
+          effect: NoExecute
+          tolerationSeconds: 300
+        - key: node.kubernetes.io/unreachable
+          operator: Exists
+          effect: NoExecute
+          tolerationSeconds: 300
+        - key: node.kubernetes.io/memory-pressure
+          operator: Exists
+          effect: NoSchedule
```
chiragjn, line 476: We should not add these

chiragjn (approved):
Approving but let's remove the not ready, unreachable, memory pressure tolerations

charts/tfy-gpu-operator/values.yaml
nikp1172: these are there in chart: like these are default values. I can remove

PR #216, opened 2024-02-26
Added affinity for notebook controller

dunefro:
1. Added support for affinity in notebook controller deployment
2. Removed default tolerations from notebook-controller-deployment.


charts/tfy-notebook-controller/templates/deployment.yaml
```diff
@@ -103,4 +103,8 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      affinity:
+        {{- with .Values.affinity }}
```
shubhamrai1993, line 107: Lets move this line above `affinity` like the previous key? This would prevent incorrect yaml in case of null affinity being supplied

shubhamrai1993 (approved):

charts/tfy-notebook-controller/templates/deployment.yaml
dunefro: By default affinity is supplied `{}`.

PR #218, opened 2024-02-29
update charts for adding default timeslicing config for azure

nikp1172:


charts/tfy-gpu-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: tfy-gpu-operator
-version: 0.1.12
+version: 0.1.13
```
chiragjn, line 3: If we are not yet finalized with values and experimenting then please keep it rc

```suggestion
version: 0.1.13-rc.1
```

chiragjn (approved):

PR #220, opened 2024-03-06
add fractional GPU config for AWS, make changes for Azure also

nikp1172:


charts/tfy-gpu-operator/values.yaml
```diff
@@ -76,6 +83,25 @@ aws-eks-gpu-operator:
...
+        value: all
+    ## @skip aws-eks-gpu-operator.devicePlugin.config
+    config:
+      data:
+        all: ''
+        time-sliced-10: |-
+          version: v1
+          sharing:
+            timeSlicing:
+              renameByDefault: true
```
chiragjn, line 98: Can we control what it gets renamed to?

nikp1172: No: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-sharing.html#configuration
It is nvidia.com/gpu.shared

chiragjn (approved):

PR #238, opened 2024-04-09
Enforced requirement for the node generation to be above 3

shubhamrai1993:


chiragjn (approved):

charts/tfy-karpenter-config/templates/controlplane/karpenter-control-plane-nodepool.yaml
```diff
@@ -36,6 +36,10 @@ spec:
           operator: In
           values:
           - linux
+        - key: karpenter.k8s.aws/instance-generation
+          operator: Gt
+          values:
+            - "3"
```
chiragjn, line 42: I am surprised K8s allows list with Gt op 😅

PR #281, opened 2024-04-24
Add a chart to manage configmaps for TrueFoundry

chiragjn:


charts/tfy-configs/templates/cicd-templates-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-cicd-templates-cm
```
chiragjn, line 4: I need help with this
Should I add `{{ .Release.Name }}` or not ?

akashg3627: this is fine

akashg3627 (approved):

PR #301, opened 2024-05-02
Put in default values as what we provide in argo application

innoavator:
This PR aims to make the default values in the chart so that we don't need to override in the argo application at https://github.com/truefoundry/infra-charts/blob/main/charts/tfy-k8s-aws-eks-inframold/templates/tfy-aws/karpenter-config.yaml.

Once this gets merged, we will reduce the configuration in the argocd application.


charts/tfy-karpenter-config/values.yaml
```diff
@@ -200,7 +202,7 @@ karpenter:
   ## Default Inferentia Node Template Config
   inferentiaDefaultNodeTemplate:
     ## @param karpenter.inferentiaDefaultNodeTemplate.enabled Enable the inferentia node pool
-    enabled: false
+    enabled: true
```
shubhamrai1993, line 205: `inferentiaNodePool` will also have to be enabled for this to take effect

charts/tfy-karpenter-config/values.yaml
```diff
@@ -54,14 +54,15 @@ karpenter:
...
+      - on-demand
     ## @param karpenter.defaultNodePool.zones Zones for the default node pool
     zones: []
     ## @param karpenter.defaultNodePool.architectures Architectures for the default node pool
     architectures:
       - amd64
     ## @param karpenter.defaultNodePool.instanceFamilies.notAllowed Not allowed instance families for the default node pool
     instanceFamilies:
-      notAllowed: ["t3", "t2", "t3a", "p2", "p3", "p4d", "p4de", "p5", "g4dn", "g5", "g4ad", "inf1", "inf2", "trn1", "trn1n"]
+      notAllowed: ["t3", "t2", "t3a", "p2", "p3", "p4d", "p4de", "p5", "g4dn", "g5", "c1", "cc1", "cc2", "cg1", "cg2", "cr1", "g1", "g2", "hi1", "hs1", "m1", "m2", "m3", "g4ad", "inf1", "inf2", "trn1", "trn1n"]
```
shubhamrai1993, line 65: I think c1 and these were removed because they have really less IP address available and frequently run into ip exhaustion issue

shubhamrai1993 (approved):

PR #328, opened 2024-05-14
feat: allowed namespaces

debajyoti-truefoundry:


charts/tfy-agent/values.yaml
```diff
@@ -435,11 +447,38 @@ tfyAgentProxy:
...
+    ## @param tfyAgentProxy.clusterRole.clusterScopedAccessRules [array] Cluster scoped access rules for the control plane.
+    ## The control plane will always have cluster-wide access to the resources below.
+    clusterScopedAccessRules:
+      - apiGroups: ["storage.k8s.io"]
+        resources: ["storageclasses"]
+        verbs: ["get", "list"]
+
+      - apiGroups: ["networking.istio.io"]
+        resources: ["virtualservices"]
+        verbs: ["list"]
```
chiragjn, line 459: Is this because we want to check against all routes for conflicts in the cluster instead of just the ones in allowed namespaces?

charts/tfy-agent/values.yaml
```diff
@@ -435,11 +447,38 @@ tfyAgentProxy:
...
+
+    ## @param tfyAgentProxy.clusterRole.namespaceScopedAccessRules [array] Namespace scoped access rules for the control plane.
+    ## If the `config.allowedNamespaces` list is not empty, the control plane can only access the namespaced resources below
+    ## if the namespace is present in `config.allowedNamespaces`.
+    ##
+    ## On the other hand, if the `config.allowedNamespaces` list is empty, the control plane will have cluster-wide access to the resources below.
+    namespaceScopedAccessRules:
+      - apiGroups: [""]
+        resources: ["namespaces"]
+        verbs: ["patch", delete]
```
chiragjn, line 477:
```suggestion
        verbs: ["patch", "delete"]
```

charts/tfy-agent/values.yaml
debajyoti-truefoundry: Yes. We will later do the same check by using the DB.

charts/tfy-agent/values.yaml
```diff
@@ -435,11 +447,38 @@ tfyAgentProxy:
...
+        resources: ["storageclasses"]
+        verbs: ["get", "list"]
+
+      - apiGroups: ["networking.istio.io"]
+        resources: ["virtualservices"]
+        verbs: ["list"]
+
+      - apiGroups: [""]
+        resources: ["nodes", "namespaces"]
+        verbs: ["list"]
```
chiragjn, line 463: No "get" here?

debajyoti-truefoundry: We do not need it.

chiragjn (approved):

chiragjn (approved):

PR #338, opened 2024-05-17
docs: update tfy agent permission doc

debajyoti-truefoundry:


chiragjn (approved):
Might need to bump the version

PR #407, opened 2024-06-11
tfy-gpu-operator: Add most allocated scoring strategy for nvidia.com/gpu

chiragjn:


chiragjn:
Will revisit in future
This might not be required considering Karpenter and NAP can do intelligent consolidation and allocation

PR #429, opened 2024-06-26
Add artifacts template generator

DeeAjayi:


scripts/upload_artifacts.py
```diff
@@ -0,0 +1,127 @@
...
+
+        logging.info(f"Pulling image: {image_url}")
+        try:
+            pulled_image = client.images.pull(image_url)
+            logging.info(f"Successfully pulled image: {image_url}")
+        except Exception as e:
+            logging.error(f"Failed to pull image: {image_url}. Error: {e}")
+            continue
+
+        image_name = image_url.split('/')[-1].split(':')[0]
```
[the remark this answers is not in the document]
DeeAjayi: That is actually most of the use-cases, it'll still work. We are only interested in the image name and tag. This logic only takes the last strings after the split which is the image name and tag regardless of the number of `/`

scripts/artifacts_template_generator.py
```diff
@@ -0,0 +1,146 @@
...
+        return None
+
+    return image_info_list
+
+if __name__ == "__main__":
+    if len(sys.argv) != 5:
+        print("Usage: python extract_and_process_chart_info.py <chart-name> <chart-repo-url> <values.yaml> <output.json>")
+        sys.exit(1)
+
+    chart_name = sys.argv[1]
```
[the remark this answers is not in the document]
DeeAjayi: Yes we ca

scripts/upload_artifacts.py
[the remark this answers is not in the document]
shubhamrai1993: I'm thinking of a situation where the image name is not unique even though the namespace is. Eg - There is a truefoundry image in dockerhub under `truefoundrycloud` and another within ecr

scripts/artifacts_template_generator.py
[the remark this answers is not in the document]
DeeAjayi: Done

scripts/artifacts_template_generator.py
```diff
@@ -0,0 +1,146 @@
...
+        manifests = yaml.safe_load_all(yaml_content)
+        image_info_list = []
+        for manifest in manifests:
+            if manifest is None:
+                continue
+            if 'kind' in manifest and 'spec' in manifest:
+                kind = manifest['kind']
+                if kind in resources_with_images:
+                    if 'template' in manifest['spec']:
+                        containers = manifest['spec']['template']['spec'].get('containers', [])
```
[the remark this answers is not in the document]
DeeAjayi: Done

scripts/upload_artifacts.py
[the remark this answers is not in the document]
DeeAjayi: What issue do you think this will cause?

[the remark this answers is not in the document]
shubhamrai1993: So two images - 
- `docker.io/truefoundrycloud/tfy-agent:latest`
- `public.ecr.aws.com/truefoundrycloud/tfy-agent:latest`
These could be completely different images since latest can be different.

When we push to our own registry, both will end up having url - 
- `<custom_registry>/tfy-agent:latest` which can be a problem

Also the kyverno patch only replaces the first part with the registry url. So `docker.io/truefoundrycloud/tfy-agent:latest` becomes `<custom_registry>/truefoundrycloud/tfy-agent:latest`

scripts/artifacts_template_generator.py
```diff
@@ -0,0 +1,165 @@
+import yaml
```
shubhamrai1993, line 1: lets also integrate this script in the github actions

scripts/artifacts/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Steps to deploy artifacts to a private registry
+
+1. Get the necessary artifacts-manifest.json file from the [infra-charts](https://github.com/truefoundry/infra-charts/tree/main/charts/) folder depending on the infrastructure you want to deploy to.
+2. Install the python dependencies
+   ```
+   pip install -r requirements.txt
+   ```
+3. The `upload_artifact.py` is a sample python scripts which handles deployment to JFrog and ECR registries, it takes in the following arguments;
+    - `artefact_type` - this takes in the artifact type which can be either `image` or `helm`
```
shubhamrai1993, line 20: We are mixing up artefact and artifact at places. Lets use artifact everywhere

.github/workflows/update-artifacts.yml
```diff
@@ -0,0 +1,54 @@
...
+        - name: Checkout Code
+          uses: actions/checkout@v3
+          with:
+            fetch-depth: 0
+
+        - name: Setup Kubectl
+          id: install-kubectl
+          uses: azure/setup-kubectl@v3
+          with:
+            version: 'latest'
```
shubhamrai1993, line 21: lets use a specific version here

.github/workflows/update-artifacts.yml
```diff
@@ -0,0 +1,54 @@
...
+          with:
+            python-version: '3.10'
+
+        - name: Install python dependencies
+          run: pip install -r scripts/artifacts/requirements.txt
+
+        # Generate artifacts manifest for inframold charts
+        - name: Generate Artifacts Manifest for Each Chart
+          run: |
+            charts_list=$(ls charts | grep inframold)
```
shubhamrai1993, line 38: lets not do this rather. this can be error prone. rather use the static values for inframold

.github/workflows/update-artifacts.yml
```diff
@@ -0,0 +1,54 @@
...
+            done
+
+        # Update the inframold artifacts manifest
+        - name: Create Pull Request
+          uses: peter-evans/create-pull-request@v6
+          with:
+            commit-message: Update Inframold artifacts manifest
+            title: [CI] Update Inframold Artifacts
+            body: Update Inframold artifacts manifest
+            branch: update-inframold-artifacts-manifest
```
shubhamrai1993, line 53: lets add a unique number to the title and branch to prevent reusing an older one if two PRs are raised simultaneously

scripts/artifacts/README.md
```diff
@@ -0,0 +1,33 @@
+# Uploading artifacts to a private registry
```
shubhamrai1993, line 1: lets create two directories `generate-artifacts-manifest` and `upload-artifacts` and group the code, requirements and associated files in these directories. right now its haphazard

scripts/upload_artifacts.py
```diff
@@ -0,0 +1,275 @@
+import docker
```
shubhamrai1993, line 1: this file needs to be deleted

scripts/artifacts_template_generator.py
```diff
@@ -0,0 +1,187 @@
...
```
shubhamrai1993, line 1: this file needs to be deleted

scripts/extra.json
```diff
@@ -0,0 +1,74 @@
+[
```
shubhamrai1993, line 1: this file needs to be deleted

scripts/requirements.txt
```diff
@@ -0,0 +1,15 @@
+boto3==1.34.144
```
shubhamrai1993, line 1: delete

scripts/values.yaml
```diff
@@ -0,0 +1,273 @@
+## @section Global Parameters
```
shubhamrai1993, line 1: what is this values.yaml?

shubhamrai1993 (approved):

PR #436, opened 2024-07-02
Fix clickhouse env

srihari-tf:


charts/tfy-llm-gateway-infra/values.yaml
```diff
@@ -222,7 +221,5 @@ clickhouse:
     storageClassName: ''
     size: 10Gi
   nats:
-    url: >-
```
akashg3627, line 225: wouldn't it be better to keep this configurable?

srihari-tf: hmm yeah can do

akashg3627 (approved):

PR #453, opened 2024-07-10
Making al2023 as the default machine type

dunefro:
1. Made `AL2023` as the default machine type for `default` and `control plane` node pools.
2. Add option for passing ami selectors to support custom AMI.


charts/tfy-karpenter-config/values.yaml
```diff
@@ -85,7 +85,7 @@ karpenter:
     # Set this to true to enable EC2 detailed cloudwatch monitoring
     detailedMonitoring: false
     ## @param karpenter.gpuDefaultNodeTemplate.amiFamily AMI family to use for node template
-    amiFamily: "AL2"
+    amiFamily: "AL2023"
```
chiragjn, line 88: we should probably add a `AL2` or `AL2023` condition here:

https://github.com/truefoundry/infra-charts/blob/49bb48f99c4e5025bdee60b4b9544ff703e7e926/charts/tfy-karpenter-config/templates/gpu/karpenter-gpu-default-ec2nodeclass.yaml#L49

shubhamrai1993 (approved):

charts/tfy-karpenter-config/values.yaml
chiragjn: AL2023 AMIs do not come with gpu drivers pre-installed

chiragjn (approved):

PR #490, opened 2024-08-05
Add imagePullSecrets for notebook controller

DeeAjayi:


charts/tfy-notebook-controller/values.yaml
```diff
@@ -14,6 +14,11 @@ labels:
 ##
 namespaceOverride: ""
 
+## Image pull secrets for the notebook controller.
+## @param imagePullSecrets Image pull secrets for the notebook controller.
```
shubhamrai1993, line 18: these are the names of the secrets?

shubhamrai1993 (approved):

PR #619, opened 2024-09-19
Adding jspolicy helm chart

shubhamrai1993:


charts/tfy-jspolicy-configs/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Parameters
+
+### replaceImageRegistry Configuration options for replacing the image registry
+
+| Name                                          | Description                                                                                                     | Value   |
+| --------------------------------------------- | --------------------------------------------------------------------------------------------------------------- | ------- |
+| `replaceImageRegistry.enabled`                | Enable or disable replacing the image registry                                                                  | `false` |
+| `replaceImageRegistry.excludeNamespaces`      | Namespaces to exclude from replacing the image registry                                                         | `[]`    |
+| `replaceImageRegistry.includeNamespaces`      | Namespaces to include from replacing the image registry. When non-empty, only these namespaces will be included | `[]`    |
```
innoavator, line 12: Which one trumps? What if I put the same namespace in include and exclude? Can you add docs for this?

innoavator (approved):

charts/tfy-jspolicy-configs/README.md
shubhamrai1993: It says this - `When non-empty, only these namespaces will be included`

PR #634, opened 2024-09-24
Fix the image and helm charts upload script

innoavator:


scripts/upload-artifacts/upload_artifacts.py
```diff
@@ -174,52 +68,49 @@ def pull_and_push_images(image_list, destination_registry, registry_type, artifa
...
+            run_command(f"helm show chart {new_chart_url}/{chart} --version {target_revision}")
+            logging.info("Chart already exists in destination repo")
+            continue
+        except Exception as e:
+            logging.error(f"Failed to download Helm chart: {chart}. Error: {e}")
+
+        # Download chart from source repo
+        logging.info(f"Downloading Helm chart: {chart} from source repo: {repo_url}")
         try:
             run_command(f"helm repo add {chart} {repo_url}")
```
shubhamrai1993, line 103: This has a bug actually. If the repo url is that of oci, then helm repo add doesn't work. I'll do the fix in a separate PR

shubhamrai1993 (approved):

PR #662, opened 2024-09-26
Update replace-image-registry.yaml to no use defaultNamespace for non…

dunefro:
… docker.io registry

### DIFF
--- a/charts/tfy-jspolicy-configs/templates/replace-image-registry.yaml
+++ b/charts/tfy-jspolicy-configs/templates/replace-image-registry.yaml
@@ -73,6 +73,10 @@ spec:
         image = image.substring(namespaceIndex + 1);
       }
 
+      // Don't use defaultNamespace for non docker.io registry
+      if (registry !== "docker.io" && namespace = defaultNamespace) {
+        namespace = '';
+      }
       // Check if there's a tag specified
       const colonIndex = image.lastIndexOf(':');
       if (colonIndex > -1) {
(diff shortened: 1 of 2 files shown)


charts/tfy-jspolicy-configs/templates/replace-image-registry.yaml
```diff
@@ -95,13 +102,14 @@ spec:
...
       let registryReplacement = registryReplacementMap[registry];
       if (!registryReplacement) {
+        if (!registryReplacementMap['*']) {
+          return imageUri;
+        }
         registryReplacement = registryReplacementMap['*'];
-      }
-      if (imageUri.startsWith(registryReplacement)) {
-        return imageUri
       } else {
```
shubhamrai1993, line 109: What is this else condition for?

dunefro: Else condition is for the case where registry is in the map. First if condition is when registry to be replaced is not in map.

shubhamrai1993: But its already assigned the same value in line #96. So this condition shouldn't be needed

charts/tfy-jspolicy-configs/templates/replace-image-registry.yaml
```diff
@@ -95,13 +102,14 @@ spec:
       let newImageUri = imageUri;
       let registryReplacement = registryReplacementMap[registry];
       if (!registryReplacement) {
+        if (!registryReplacementMap['*']) {
+          return imageUri;
+        }
         registryReplacement = registryReplacementMap['*'];
-      }
-      if (imageUri.startsWith(registryReplacement)) {
```
shubhamrai1993, line 100: Why remove this check? This was needed to prevent recursive replacement

shubhamrai1993 (approved):

PR #670, opened 2024-10-01
add python version to image tags

nikp1172:


charts/tfy-configs/files/workbench-images.yaml
chiragjn, line 1: Bump the chart version please

chiragjn (approved):
Approved, but needs chart version bump

PR #672, opened 2024-10-03
feat(charts): add support for imagePullSecrets in device plugin, scheduler extension, and scheduler to allow pulling images from private repositories

sachincool:


charts/tfy-inferentia-operator/values.yaml
```diff
@@ -3,6 +3,8 @@
 ## @param devicePlugin.resources.requests.cpu CPU request for device plugin Daemonset.
 ## @param devicePlugin.resources.requests.memory Memory request for device plugin Daemonset.
 ## @param devicePlugin.image Image to use for device plugin Daemonset.
+imagePullSecrets: []
+## @param imagePullsecrets List of image pull secrets
```
DeeAjayi, line 7: The comment should come before the key, else the README generator won't work

DeeAjayi (changes requested):

charts/tfy-inferentia-operator/values.yaml
```diff
@@ -65,58 +67,16 @@ scheduler:
       cpu: 100m
       memory: 50Mi
   tolerations:
-      ## @skip scheduler.tolerations[0]
+    ## @skip scheduler.tolerations[0]
     - key: CriticalAddonsOnly
       operator: Exists
-  affinity:
```
debajyoti-truefoundry, line 71: Why is this removed?

debajyoti-truefoundry: oh okay, i do not think the scheduler needs to run on an inf instance

debajyoti-truefoundry: We only need it for the device plugin.

charts/tfy-inferentia-operator/templates/neuron-device-plugin.yaml
```diff
@@ -42,6 +42,10 @@ kind: ServiceAccount
 metadata:
   name: {{ .Release.Name }}-neuron-device-plugin
   namespace: {{ .Release.Namespace }}
+{{- if .Values.imagePullSecrets }}
```
shubhamrai1993, line 45: If we have two different keys for imagePullSecrets, we should perform the override before using the service specific one here or in the pod spec. We don't really need to inject at the service account level tbh

shubhamrai1993: This is a good example https://github.com/argoproj/argo-helm/blob/d61432be3c1e460e7037462b6b5d5f8f1bdcc6d7/charts/argo-cd/templates/argocd-server/deployment.yaml#L47

sachincool: Thanks, I made the changes

charts/tfy-inferentia-operator/templates/neuron-scheduler.yaml
```diff
@@ -122,6 +122,10 @@ spec:
         version: second
     spec:
       priorityClassName: system-cluster-critical
+      {{- with .Values.scheduler.imagePullSecrets | default .Values.imagePullSecrets }}
```
DeeAjayi, line 125: there is no reference to .Values.scheduler.imagePullSecrets in the values.yaml file. It can be an empty list mainly for reference purpose

sachincool: have made the changes, added those as well.

sachincool: affinity docs were missing, have added those as well.

DeeAjayi (approved):

PR #674, opened 2024-10-04
Making whenUnsatisfiable to ScheduleAnyway so that nats pods always gets scheduled

dunefro:
1. Ensuring nats pods are scheduled always.
2. Moving tfy-llm-gateway-infra to `v0.1.0`


charts/tfy-llm-gateway-infra/values.yaml
```diff
@@ -195,7 +195,7 @@ nats:
     topologySpreadConstraints:
       kubernetes.io/hostname:
```
DeeAjayi, line 196: `kubernetes.io/hostname` a valid key for `topologySpreadConstraints` - https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/

srihari-tf (approved):

PR #736, opened 2024-10-25
Pushing helm charts to artifactory

dunefro:
1. Created a workflow to push helm charts to truefoundry public artifactory.
2. Create a python script to support the above


scripts/artifactory-chart-releaser/main.py
```diff
@@ -0,0 +1,64 @@
...
+    
+    package_file = f"{chart_name}-{chart_version}.tgz"
+    try:
+        subprocess.run(
+            ["helm", "push", package_file, f"oci://{ARTIFACTORY_REPOSITORY_URL}"],
+            check=True,
+            cwd=chart_dir
+        )
+        print(f"Pushed helm chart {package_file} to the OCI repository oci://{ARTIFACTORY_REPOSITORY_URL}")
+    except subprocess.CalledProcessErro as e:
```
DeeAjayi, line 49:
```suggestion
    except subprocess.CalledProcessError as e:
```

DeeAjayi (changes requested):

scripts/artifactory-chart-releaser/main.py
dunefro: Should I just keep it like `except Exception as e:`

DeeAjayi: subprocess.CalledProcessError is fine. that is the error raised when subprocess.run() is called and check=True. - https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError

> If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured. ld the arguments, the exit code, and stdout and stderr if they were captured. )

DeeAjayi (approved):

PR #787, opened 2024-11-13
update resource + remove rate limit config

srihari-tf:


charts/tfy-llm-gateway/values.yaml
```diff
@@ -142,20 +142,13 @@ env:
...
-
 ## Autoscaling configuration
 autoscaling:
   ## @param autoscaling.enabled Enable autoscaling  
   enabled: false
   ## @param autoscaling.minReplicas Minimum number of replicas
-  minReplicas: 1
+  minReplicas: 3
   ## @param autoscaling.maxReplicas Maximum number of replicas
   maxReplicas: 10
```
innoavator, line 152: Make this number much higher - like 100 maybe

innoavator (dismissed):

PR #846, opened 2024-12-04
Add Extra tags to karpenter config

DeeAjayi:


sachincool (approved):

charts/tfy-karpenter-config/templates/controlplane/karpenter-control-plane-ec2nodeclass.yaml
```diff
@@ -23,6 +23,8 @@ spec:
         {{ $key }}: {{ $val | quote }} 
         {{- end }}
   tags:
+    cpu: "true"
```
sachincool, line 26: we should rely on a tag, which can take both values. and then use filter on top, 
example: `compute_type: cpu|gpu`

shubhamrai1993: +1

charts/tfy-karpenter-config/templates/default/karpenter-default-ec2nodeclass.yaml
```diff
@@ -22,6 +22,8 @@ spec:
         {{ $key }}: {{ $val | quote }} 
         {{- end }}
   tags:
+    cpu: "true"
```
shubhamrai1993, line 25: can't we just use the extraTags field already present and only make the change in our env. Hardcoding the tags increases the chance of conflict with existing setup at customer's end

shubhamrai1993 (approved):

charts/tfy-karpenter-config/templates/default/karpenter-default-ec2nodeclass.yaml
shubhamrai1993: another option is to use truefoundry.com/compute_type as the tag key

shubhamrai1993 (approved):

PR #868, opened 2024-12-12
chore: missing services/status resource permission

rushHub46:
To allow access to sub resource `services/status` so that readNamespacedServiceStatus() does not throw a permission issue during cluster sync.


rushHub46:
Letting this PR be closed for now as this did not seem the correct solution to solving the syncCluster API which failed when calling k8s-client API.

PR #891, opened 2024-12-20
Add checks for truefoundry images multi-arch compatibility

DeeAjayi:
Add a check if truefoundry charts control plane components support both ARM and AMD platforms


scripts/generate-artifacts-manifest/check_chart_images_support_multiple_architectures.py
```diff
@@ -0,0 +1,96 @@
...
+        logging.error(f"Whitelisted images file not found at {whitelisted_images_path}")
+        sys.exit(1)
+
+    return artifacts_manifest_path, whitelisted_images_path
+
+
+def get_truefoundry_chart_images(artifacts_manifest):
+    """Extract the list of images associated with the 'truefoundry' chart."""
+    for artifact in artifacts_manifest:
+        if artifact.get("details", {}).get("chart") == "truefoundry":
```
shubhamrai1993, line 47: We need to include all truefoundry built charts. So this would include `truefoundry`, tfy-agent, tfy-notebook-controller, elasti etc

sachincool (dismissed):

sachincool (dismissed):

shubhamrai1993 (approved):

PR #897, opened 2024-12-24
[jspolicy config] fix pod volume mounts for configMap

akashg3627:


shubhamrai1993 (approved):

charts/tfy-jspolicy-configs/templates/pod-volume-mounts.yaml
```diff
@@ -78,6 +78,12 @@ spec:
           print("Invalid volume mount details: " + JSON.stringify(mountDetails));
           exit();
         }
+        // skip if the volume mount already exists
+        if (container.volumeMounts.find(volumeMount => volumeMount.name === volumeName)) {
```
chiragjn, line 82: Future TODO: allowing multiple volumeMounts from the same volume as long as mountPath and subPath are not same

chiragjn (approved):

charts/tfy-jspolicy-configs/templates/pod-volume-mounts.yaml
```diff
@@ -38,13 +38,13 @@ spec:
     // add the volumes to the pod
     const podVolumes = request.object.spec.volumes || [];
     mountDetails.forEach(mountDetails => {
```
chiragjn, line 40: We should rename the each item - both the list and item are being called `mountDetails`

This can cause confusion

PR #899, opened 2024-12-26
Add sds server to agent deployment

gowthamvadisetti:


charts/tfy-agent/templates/tfy-agent-deployment.yaml
```diff
@@ -38,6 +38,8 @@ spec:
             - name: CONTROL_PLANE_NATS_URL
               value: {{ .Values.config.controlPlaneNatsClusterIP | quote }}
             {{- end }}
+            - name: SDS_SERVER_URL
+              value: {{ printf "%s.%s.svc.cluster.local:8000" .Release.Name .Release.Namespace | quote }}
```
nikp1172, line 42: will this not need http?

nikp1172: like we are passing host and not utl....

gowthamvadisetti: checked this one, we are communicating with sds server using grpc so adding http:// would be misleading as well

ideally we should have used a differerent name instead SDS_SERVER_URL or sdsServerURL we need to update everywhere as well

charts/tfy-agent/values.yaml
```diff
@@ -454,3 +454,93 @@ resourceQuota:
   ## @param resourceQuota.priorityClasses PriorityClasses to enable.
   priorityClasses:
     - system-cluster-critical
+
+
+## Config for SDS service
+##
+sdsServer:
+  ## @param sdsServer.enabled Bool value to deploy sdsServer
+  enabled: false
```
nikp1172, line 463: make default true

nikp1172 (approved):

charts/tfy-agent/templates/tfy-agent-deployment.yaml
nikp1172: fine, we can keep it like this

charts/tfy-agent/templates/sds-deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          env:
+            {{- if .Values.config.controlPlaneURL }}
+            - name: CONTROL_PLANE_URL
+              value: {{ .Values.config.controlPlaneURL | quote }}
+            {{- else }}
+            - name: CONTROL_PLANE_URL
+              value: {{ .Values.config.controlPlaneClusterIP | quote }}
+            {{- end }}
```
nikp1172, line 52: why are we doing this? just controlPlaneUrl should be fine

nikp1172 (approved):

PR #901, opened 2024-12-30
chore: allow agent to access gateways

rushHub46:


charts/tfy-agent/templates/tfy-agent-proxy-clusterrole-cs.yaml
```diff
@@ -20,8 +20,8 @@ rules:
     verbs: ["list"]
 
   - apiGroups: ["networking.istio.io"]
-    resources: ["virtualservices"]
-    verbs: ["list"]
+    resources: ["virtualservices", "gateways"]
+    verbs: ["list", "get"]
```
akashg3627, line 24: get is not needed, list should be enough for our case

akashg3627, line 24:
```suggestion
    verbs: ["list"]
```

akashg3627 (approved):

PR #911, opened 2025-01-03
refactor: keep schema config explicitly in the values file.

debajyoti-truefoundry:
This will allow us to move to tsdb in future.

### DIFF
--- a/charts/tfy-loki/values.yaml
+++ b/charts/tfy-loki/values.yaml
@@ -4,6 +4,16 @@ loki:
   ## @param loki.enabled Enable loki
   enabled: true
   loki:
+    ## @skip loki.schemaConfig
+    schemaConfig:
+      configs:
+        - from: 2022-01-11
+          store: boltdb-shipper
+          object_store: filesystem
+          schema: v12
+          index:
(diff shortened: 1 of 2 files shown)


charts/tfy-loki/values.yaml
```diff
@@ -4,6 +4,16 @@ loki:
   ## @param loki.enabled Enable loki
   enabled: true
   loki:
+    ## @skip loki.schemaConfig
```
debajyoti-truefoundry, line 7: https://github.com/grafana/loki/blob/111c19b38147837041c6cc7c7483abe00c0c1599/production/helm/loki/values.yaml#L161C1-L168C24

charts/tfy-loki/values.yaml
```diff
@@ -4,6 +4,16 @@ loki:
...
+    schemaConfig:
```
debajyoti-truefoundry, line 8: https://github.com/grafana/loki/blob/111c19b38147837041c6cc7c7483abe00c0c1599/production/helm/loki/values.yaml#L161C1-L168C24

chiragjn (approved):

PR #912, opened 2025-01-04
Karpenter config upgrade to v1

dunefro:
Upgrade of karpenter-config to v1 webhook


charts/tfy-karpenter-config/values.yaml
```diff
@@ -20,27 +20,28 @@ karpenter:
     instanceProfile: ""
     ## @param karpenter.defaultNodeTemplate.rootVolumeSize Size for the root volume attached to node
     rootVolumeSize: 100Gi
-    ## @param karpenter.defaultNodeTemplate.extraTags Additional tags for the node template.
+    ## @param karpenter.defaultNodeTemplate.extraTags [object] Additional tags for the node template.
     extraTags: {}
     ## @param karpenter.defaultNodeTemplate.amiFamily AMI family to use for node template
-    amiFamily: "AL2023"
```
shubhamrai1993, line 26: Whats happening with amiFamily?

charts/tfy-karpenter-config/values.yaml
```diff
@@ -49,26 +50,29 @@ karpenter:
...
-    ## @param karpenter.defaultNodePool.taints Taints for the default node pool
+    ## @param karpenter.defaultNodePool.taints [array] Taints for the default node pool
     taints: []
-    ## @param karpenter.defaultNodePool.disruption.consolidationPolicy Consolidation policy for disruption
+    ## @param karpenter.defaultNodePool.disruption [object] Consolidation policy for disruption
     disruption:
-      consolidationPolicy: WhenUnderutilized
-    ## @param karpenter.defaultNodePool.capacityTypes Capacity types for the default node pool
+      budgets:
+      - nodes: 10%
```
shubhamrai1993, line 60: What happens if there are just 4 nodes for example? Will this allow disruption of 1 node?

charts/tfy-karpenter-config/values.yaml
```diff
@@ -49,26 +50,29 @@ karpenter:
...
+      consolidateAfter: 0s
```
shubhamrai1993, line 61: Should we set something like 2m here? Might help reduce the churn

charts/tfy-karpenter-config/values.yaml
dunefro: It can still be used but will cause drift of all nodes

charts/tfy-karpenter-config/values.yaml
dunefro: Yes.

shubhamrai1993 (approved):

PR #922, opened 2025-01-07
Add installation with pre-requisites

sachincool:
- **feat(pre-requisites.sh): add automated tool installation and OS/arch detection**
- **refactor(pre-requisites.sh): enhance tool installation and system detection**

### DIFF
--- a/scripts/infra-automation/pre-requisites.sh
+++ b/scripts/infra-automation/pre-requisites.sh
@@ -127,7 +127,7 @@ install_essential_utilities() {
         echo -e "${YELLOW}Install missing utilities? (Y/n)${NC}"
         read -r response
         if [[ ! $response =~ ^[nN] ]]; then
-            log_debug "Installing utilities... ${missing_tools[@]}"
+            log_debug "Installing utilities... ${missing_tools[*]}"
             case $PACKAGE_MANAGER in
                 apt-get) 
                     run_with_sudo apt-get update -qq >/dev/null 2>&1
@@ -539,6 +539,9 @@ display_installed_versions() {
         echo -e "\n${BLUE}$cloud_provider Tools:${NC}"
         read -r -a cloud_tools <<< "$(get_cloud_tools "$cloud_provider")"
         for tool in "${cloud_tools[@]}"; do
+					if [ "$tool" == "gke-gcloud-auth-plugin" ]; then
+						continue
+					fi
             if command_exists "$tool"; then
                 local version
                 version=$(get_tool_version "$tool")
--- a/scripts/infra-automation/pre-requisites.sh
+++ b/scripts/infra-automation/pre-requisites.sh
@@ -22,9 +22,13 @@ declare -A TOOL_VERSIONS=(
 declare OS PACKAGE_MANAGER ARCH HAS_SUDO
 
 # Logging functions
-log() { echo -e "${2}[TF] $3${NC}"; }
+log() { echo -e "${2} $3${NC}"; }
 log_info() { log "INFO" "$BLUE" "$1"; }
-log_debug() { log "DEBUG" "$YELLOW" "$1"; }
+log_debug() { 
+    if  [ "$TF_DEBUG" = true ]; then 
+        log "DEBUG" "$YELLOW" "$1"; 
+    fi
+}
 log_success() { log "SUCCESS" "$GREEN" "$1"; }
 log_error() { log "ERROR" "$RED" "$1"; }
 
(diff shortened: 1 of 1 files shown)


sachincool:
[asciicast](https://asciinema.org/a/4xj2ed0crtLo7fbLJL9ZkDGxx)

scripts/infra-automation/pre-requisites.sh
```diff
@@ -1,395 +1,670 @@
...
+            HAS_SUDO=true
+        fi
+    fi
+    HAS_SUDO=false
 }
 
-function log_success() {
-	echo -e "${GREEN}pre-requisite.sh - [INFO] - $1${NC}"
+# Function to run command with sudo if available
+run_with_sudo() {
```
shubhamrai1993, line 46: shouldn't we fail the whole run if sudo is not available instead of trying to run without it?

scripts/infra-automation/pre-requisites.sh
```diff
@@ -1,395 +1,670 @@
...
-# Function to log info
-function log_debug() {
-	echo -e "${YELLOW}pre-requisite.sh - [DEBUG] - $1${NC}"
+# System information
+declare OS PACKAGE_MANAGER ARCH HAS_SUDO
+
+# Logging functions
+log() { echo -e "${2}[TF] $3${NC}"; }
+log_info() { log "INFO" "$BLUE" "$1"; }
+log_debug() { log "DEBUG" "$YELLOW" "$1"; }
```
shubhamrai1993, line 27: lets add a way to not print these logs unless an env variable is present

scripts/infra-automation/pre-requisites.sh
```diff
@@ -1,395 +1,670 @@
...
-        echo "${INSTALL_URLS[$idx]}"
+# System detection
+detect_system() {
+    check_sudo
+    log_debug "System check: sudo=$HAS_SUDO"
+
+    if [[ "$OSTYPE" == "linux-gnu"* || "$OSTYPE" == "linux-musl"* ]]; then
+        OS="linux"
+        for pm in apt-get yum dnf apk; do
+            if command_exists "$pm"; then
```
shubhamrai1993, line 91: what would happen in else here?

scripts/infra-automation/pre-requisites.sh
```diff
@@ -1,395 +1,670 @@
...
+    local missing_tools=()
+    local tools_list
+    tools_list=$(get_essential_tools "$PACKAGE_MANAGER")
+    
+    for tool in $tools_list; do
+        command_exists "$tool" || missing_tools+=("$tool")
+    done
+    if [ ${#missing_tools[@]} -ne 0 ]; then
+        log_info "Missing utilities: ${missing_tools[*]}"
+        echo -e "${YELLOW}Install missing utilities? (Y/n)${NC}"
```
shubhamrai1993, line 127: shouldn't we ask this per utility?

scripts/infra-automation/pre-requisites.sh
sachincool: this is for container environment, example ubuntu:latest where sudo binaries are not installed, but you can run as root user.

scripts/infra-automation/pre-requisites.sh
sachincool: I did try that, but I liked the UX of blue/yellow logs for info and debug status. 
right now debug logs are just as important as info, but just with the color contrast change for importance. 
without the debug logs or this contrast the commands were lacking life.

scripts/infra-automation/pre-requisites.sh
sachincool: we're asking for small utilities in a single go for better UX. doing a lot of confirmations for this will bring a bad experience.

scripts/infra-automation/pre-requisites.sh
sachincool: `[[ -z $PACKAGE_MANAGER ]] && { log_error "No supported package manager found"; exit 1; }`
        
        it will break.

scripts/infra-automation/pre-requisites.sh
shubhamrai1993: what happens on user machines where the executing user is not allowed to use sudo

scripts/infra-automation/pre-requisites.sh
shubhamrai1993: from user perspective it is unlikely they will be interested in debug level logs. we need to be deliberate in exactly what we want to show to the user. rest everything can be logged like any normal service but silenced by default

scripts/infra-automation/pre-requisites.sh
shubhamrai1993: lets do the same for brew?

scripts/infra-automation/pre-requisites.sh
sachincool: I tried it on cloud shells, which have all major tools, but versions are a little behind. It just complains and exists
image

for other tools which can be downloaded we move them in path

shubhamrai1993 (dismissed):

shubhamrai1993 (approved):

PR #937, opened 2025-01-09
feat: add a chart to monitor prometheus

debajyoti-truefoundry:


akashg3627 (approved):

charts/tfy-grafana/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: tfy-grafana
 description: Truefoundry grafana provisioner
-version: 0.1.6
+version: 0.1.7
```
chiragjn, line 4: Is the dashboard final? Maybe uae rc versions till then

chiragjn (approved):

charts/tfy-grafana/Chart.yaml
debajyoti-truefoundry: good point. done.

chiragjn (approved):

PR #944, opened 2025-01-10
Fix setting of SDS server env

gowthamvadisetti:


charts/tfy-agent/templates/tfy-agent-deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
             {{- end }}
             {{- if .Values.sdsServer.enabled}}
             - name: SDS_SERVER_URL
-              value: {{ printf "%s.%s.svc.cluster.local:8000" .Release.Name .Release.Namespace | quote }}
+              value: {{ printf "sds-server.%s.svc.cluster.local:8000" .Release.Namespace | quote }}
```
akashg3627, line 43: this is also not correct

gowthamvadisetti: correct only no? The issue i had got fixed with this change

akashg3627: we should have release-name - in the name

charts/tfy-agent/templates/tfy-agent-deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
             {{- end }}
             {{- if .Values.sdsServer.enabled}}
             - name: SDS_SERVER_URL
-              value: {{ printf "%s.%s.svc.cluster.local:8000" .Release.Name .Release.Namespace | quote }}
+              value: {{ printf "%s-%s.svc.cluster.local:8000" (include "sds-server.fullname" .) .Release.Namespace | quote }}
```
akashg3627, line 43:
```suggestion
              value: {{ printf "%s.%s.svc.cluster.local:8000" (include "sds-server.fullname" .) .Release.Namespace | quote }}
```

akashg3627 (approved):

PR #951, opened 2025-01-13
Update tfy-agent permissions to create/read/delete SparkApplication

chiragjn:


charts/tfy-agent/templates/tfy-agent-proxy-clusterrole-ns.yaml
```diff
@@ -70,13 +70,24 @@ rules:
...
+
+  {{/* This allows us to apply VirtualService for Spark UI when applying a SparkApplication (Spark Job Run) */}}
+  - apiGroups: ["networking.istio.io"]
+    resources: ["virtualservices"]
+    verbs: ["create", "get", "patch"]
+
+  {{/* This allows us to apply/delete SparkApplication (Spark Job Run) */}}
+  - apiGroups: ["sparkoperator.k8s.io"]
+    resources: ["sparkapplications"]
+    verbs: ["create", "get", "delete"]
```
debajyoti-truefoundry, line 90: Just for my understanding, why do we need `get`?

chiragjn: I'll remove it for now, I was thinking we might need it we want to get the driver pod uid from status field and patch the VS

charts/tfy-agent/templates/tfy-agent-proxy-clusterrolebinding-ns.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.tfyAgentProxy.enabled -}}
 {{- if .Values.tfyAgentProxy.clusterRole.enable -}}
-{{- if or .Values.config.allowedNamespaces .Values.tfyAgentProxy.clusterRole.strictMode }}
```
chiragjn, line 3: I removed this because line 4 is `not allowedNamespaces`

PR #961, opened 2025-01-15
fix: missing services status permission for proxy service account

rushHub46:


charts/tfy-agent/templates/tfy-agent-proxy-clusterrole-ns.yaml
```diff
@@ -27,7 +27,7 @@ rules:
     verbs: ["delete", "get"]
 
   - apiGroups: [""]
-    resources: ["pods/log", "events", "services", "persistentvolumes"]
+    resources: ["pods/log", "events", "services", "services/status", "persistentvolumes"]
```
akashg3627, line 30: as we do not need list, you can add this as another section with just get verb

dunefro (approved):

chiragjn (dismissed):

chiragjn (approved):

PR #1033, opened 2025-02-11
feat(values.yaml): add labels and taints for control plane node pool to enhance node identification and scheduling refactor(values.yaml): remove spot capacity type from control plane node pool to ensure consistent resource availability feat(values.yaml): introduce providers section with default-on-demand to specify capacity type requirements

sachincool:


dunefro (commented):
Please update Chart.yaml

dunefro (approved):

PR #1089, opened 2025-02-27
update default values and limits for helm charts

Maanas-23:


charts/tfy-karpenter/values.yaml
```diff
@@ -30,8 +30,8 @@ karpenter:
   controller:
     resources:
       requests:
-        cpu: 50m
-        memory: 100Mi
+        cpu: 0.56
```
shubhamrai1993, line 33: lets sanitise these values. Make it 0.5 for request and 1 for limit should be fine

shubhamrai1993 (approved):

shubhamrai1993 (approved):

shubhamrai1993 (approved):

PR #1118, opened 2025-03-06
Add labels and annotations to tfy-llm-gateway-infra

DeeAjayi:


charts/tfy-llm-gateway-infra/templates/clickhouse-installation.yaml
```diff
@@ -5,6 +5,11 @@ metadata:
   name: {{ .Release.Name }}
   annotations:
     argocd.argoproj.io/sync-options: Prune=false,Delete=false
```
dunefro, line 7: We should move this in the helpers.tpl

charts/tfy-llm-gateway-infra/templates/clickhouse-installation.yaml
```diff
@@ -5,6 +5,11 @@ metadata:
...
+    {{- if include "clickhouse.annotations" . -}}
```
dunefro, line 8: We should be removing this

charts/tfy-llm-gateway-infra/values.yaml
```diff
@@ -34,6 +34,12 @@ global:
       ## clickhouse password
       ##
       passwordKey: "CLICKHOUSE_USER_PASSWORD"
+  ## @param global.labels [object] labels
+  ##
+  labels: {}
+  ## @param global.annotations [object] annotations
+  ##
+  annotations: {}
```
dunefro, line 42: Let's add support for nats.labels and nats.annotations also in the values.yaml so that we can reference them in the helpers.tpl

charts/tfy-llm-gateway-infra/templates/_helper.tpl
```diff
@@ -0,0 +1,47 @@
...
+{{- end }}
+
+{{/*
+nats annotations
+*/}}
+{{- define "nats.annotations" -}}
+{{- if .Values.global.annotations }}
+{{- toYaml .Values.global.annotations }}
+{{- else }}
+{}
```
dunefro, line 45: We should also check for nats annotations

charts/tfy-llm-gateway-infra/templates/_helper.tpl
```diff
@@ -0,0 +1,47 @@
...
+{{- define "clickhouse.virtualservice.annotations" -}}
+{{- if .Values.clickhouse.annotations }}
+{{- toYaml .Values.clickhouse.annotations }}
+{{- else if .Values.global.annotations }}
+{{- toYaml .Values.global.annotations }}
+{{- else }}
+{}
+{{- end }}
+{{- end }}
+
```
dunefro, line 37: Where is "nats.labels" - is this chart even working ?

charts/tfy-llm-gateway-infra/templates/nats-metrics-virtualservice.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
 metadata:
   name: "{{ .Release.Name }}-nats-metrics"
+  labels:
+    {{- include "nats.labels" . | nindent 4 }}
```
dunefro, line 7: Where is this function ?

charts/tfy-llm-gateway-infra/templates/_helper.tpl
DeeAjayi: There is a nats.labels in the nats charts dependency which was conflicting with the one in the helpers function. This would reference the existing one.

charts/tfy-llm-gateway-infra/templates/nats-metrics-virtualservice.yaml
DeeAjayi: this is in the nats dependency chart

dunefro (approved):

PR #1128, opened 2025-03-10
fix: OOM Kill alert considers container restart

ttanay:
The `kube_pod_container_status_last_terminated_reason` gives us the reasons of the last terminated container.

If the container restarts due to OOM Kill but then runs succesfully, the metric will continue to report the OOM Kill for the same pod.

Therefore, we need to also consider whether a container restart happened in the last 5 minutes.

The ideal solution is to use the metric -- `container_oom_events_total`, but it reports 0 always[1]. 

1. https://github.com/google/cadvisor/issues/3015

### DIFF
--- a/charts/tfy-prometheus-config/templates/prometheusRules/alerting-rules.yaml
+++ b/charts/tfy-prometheus-config/templates/prometheusRules/alerting-rules.yaml
@@ -56,6 +56,8 @@ spec:
     - for: 2m
       expr: >-
         sum by (container, pod, namespace) (sum_over_time(kube_pod_container_status_last_terminated_reason{reason="OOMKilled", job="kube-state-metrics"}[5m])) > 0
+        and
+        max by (container, pod, namespace) (increase(kube_pod_container_status_restarts_total[5m]) > 0)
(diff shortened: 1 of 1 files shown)


charts/tfy-prometheus-config/templates/prometheusRules/alerting-rules.yaml
```diff
@@ -56,6 +56,8 @@ spec:
     - for: 2m
       expr: >-
         sum by (container, pod, namespace) (sum_over_time(kube_pod_container_status_last_terminated_reason{reason="OOMKilled", job="kube-state-metrics"}[5m])) > 0
+        and
```
shubhamrai1993, line 59: shouldn't we be joining the result of the two queries?

ttanay: > also why do we need `sum_over_time`?

This was being used in the alert summary. But, this was incorrect. 
Removed it as the count is not significant. The fact that it is being OOM Killed is.

shubhamrai1993 (approved):

PR #1135, opened 2025-03-12
Add ServiceMonitor for TrueFoundry K8s Controller

GauravWaghmare:
Create a new ServiceMonitor template for monitoring the TrueFoundry Kubernetes Controller with configurable options


charts/tfy-prometheus-config/values.yaml
```diff
@@ -806,6 +806,29 @@ serviceMonitors:
     namespaceSelector:
       matchNames:
         - truefoundry
+  # serviceMonitors.tfyK8sController Service monitor for tfy-k8s-controller
+  tfyK8sController:
+    ## @param serviceMonitors.tfyK8sController.enabled Enable service monitor for tfy-k8s-controller
+    enabled: true
```
shubhamrai1993, line 812: lets keep it disabled by default. this is only relevant for control plane clusters

shubhamrai1993:
We also need to bump the version in Chart.yaml for `tfy-prometheus-config` for the new version of the chart to get published

charts/tfy-prometheus-config/values.yaml
GauravWaghmare: Done

shubhamrai1993 (approved):

PR #1153, opened 2025-03-18
Update tfy-k8s-controller service name to http-3002

GauravWaghmare:
Update port name to match the port name set by sfy-server


charts/truefoundry/templates/tfy-k8s-controller/service.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     - port: {{ .Values.tfyK8sController.service.port }}
       targetPort: {{ .Values.tfyK8sController.service.port }}
       protocol: TCP
-      name: http
+      name: http-3002
```
shubhamrai1993, line 16: lets use the port that is present in values file. similar to tfy-controller

GauravWaghmare: Done. Thanks.

PR #1165, opened 2025-03-20
Added support for flexible requirements

dunefro:
1. Bug fix for limits in default and control plane nodepools
2. Adding support for limits in all node pools
3. Adding support for additional requirements in all nodepools.
4. Add support for metadata options in ec2nodeclass

### DIFF
--- a/charts/tfy-karpenter-config/values.yaml
+++ b/charts/tfy-karpenter-config/values.yaml
@@ -44,7 +44,7 @@ karpenter:
     #   httpPutResponseHopLimit: 1
     #   httpTokens: required
     metadataOptions: {}
-    ## @param karpenter.defaultNodeTemplate.detailedMonitoring. Set this to true to enable EC2 detailed cloudwatch monitoring.
+    ## @param karpenter.defaultNodeTemplate.detailedMonitoring Set this to true to enable EC2 detailed cloudwatch monitoring.
     detailedMonitoring: false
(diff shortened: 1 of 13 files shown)


chiragjn (approved):

charts/tfy-karpenter-config/values.yaml
```diff
@@ -77,11 +82,10 @@ karpenter:
     ## @param karpenter.defaultNodePool.instanceSizes.notAllowed [array] Not allowed instance sizes for the default node pool
     instanceSizes:
       notAllowed: ["nano", "micro", "small", "medium", "large", "metal", "metal-16xl", "metal-24xl", "metal-48xl"]
-    ## @param karpenter.defaultNodePool.limits.cpu CPU limit for the default node pool
-    limits:
-      cpu: 75
-      ## @param karpenter.defaultNodePool.limits.memory Memory limit for the default node pool
-      memory: 300Gi
+    ## @param karpenter.defaultNodePool.limits [object] compute limits for the default nodepool
+    limits: {}
```
shubhamrai1993, line 86: Why remove limits completely? These can always be overridden

dunefro: These were never used as well. If you see we were using 
```
  {{- if .Values.karpenter.gpuNodePool.limits }}
  limits:
  {{- .Values.karpenter.gpuNodePool.limits | toYaml | nindent 4 }}
  {{- end }}
```
in default and control plane nodepool which is completely incorrect

PR #1229, opened 2025-04-04
make clickhouse run with non-root user

akashg3627:


charts/tfy-clickhouse/templates/clickhouse-installation.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         <clickhouse>
             <logger>
                 <level>information</level>
-                <log>/var/log/clickhouse-server/clickhouse-server.log</log>
-                <errorlog>/var/log/clickhouse-server/clickhouse-server.err.log</errorlog>
+                <log>/tmp/log/clickhouse-server/clickhouse-server.log</log>
+                <errorlog>/tmp/log/clickhouse-server/clickhouse-server.err.log</errorlog>
```
debajyoti-truefoundry, line 36: Can we pipe the logs to stdio?

charts/tfy-clickhouse/templates/clickhouse-installation.yaml
```diff
@@ -91,6 +91,17 @@ spec:
...
+            - name: tmp
+              emptyDir: {}
+          securityContext:
+            capabilities:
+              add:
+                - CAP_NICE
+                - CAP_IPC_LOCK
+            fsGroup: 101
+            runAsGroup: 101
+            runAsUser: 101
```
debajyoti-truefoundry, line 104: This can potentially create an issue with the existing data on EBS.
Please test on Devtest.

debajyoti-truefoundry: Check if Clickhouse can load the existing data once you apply this change.

chiragjn: Let's stick with `1000` as that is pretty common unless we have good reasons for 101

PR #1230, opened 2025-04-04
chore(istio): update Istio API version to v1 and upgrade Istio dependencies

sachincool:
Update all Istio-related resources to use `networking.istio.io/v1`
API version for better compatibility and future-proofing. Upgrade
Istio dependencies and image tags from 1.21.1 to 1.25.1 to leverage
new features and improvements. Add `tfy-gateway-api-crd` dependency
to ensure compatibility with the updated Istio version.


charts/tfy-istio-ingress/Chart.yaml
```diff
@@ -1,10 +1,13 @@
 apiVersion: v2
 name: tfy-istio-ingress
-version: 0.1.4
+version: 0.1.5
```
dunefro, line 3: Let's make it `0.2.0`

PR #1234, opened 2025-04-07
istio-ingress updates

dunefro:


charts/tfy-istio-ingress/values.yaml
```diff
@@ -1,41 +1,26 @@
...
-  ## @param gateway.autoscaling.minReplicas Minimum number of replicas for autoscaling the Gateway.
-  ##
   autoscaling:
+    ## @param gateway.autoscaling.minReplicas Minimum number of replicas for autoscaling the Gateway.
+    ##
     minReplicas: 3
-  ## Service configuration for the Gateway.
+    ## @param gateway.autoscaling.maxReplicas Maximum number of replicas for autoscaling the Gateway.
+    ##
+    maxReplicas: 10
```
innoavator, line 10: This should be higher? Something like 100?

PR #1297, opened 2025-04-21
Add global affinity and tolerations to truefoundry chart dependencies

DeeAjayi:


charts/tfy-buildkitd-service/templates/_helpers.tpl
```diff
@@ -196,4 +196,30 @@ limits:
...
 200Gi
 {{- end }}
+{{- end }}
+
+{{/*
+Affinity for the buildkitd service
+*/}}
+{{- define "buildkitd-service.affinity" -}}
+{{- if .Values.affinity -}}
+{{ toYaml .Values.affinity }}
```
dunefro, line 206: If we are using `nindent` in the main template file, should we make it
```
{{- toYaml
```

dunefro (commented):
Please bump up the chart versions as well

charts/tfy-buildkitd-service/templates/_helpers.tpl
dunefro: If yes, we should make the change everywhere

DeeAjayi: is there a significant difference or advantage to doing this? I already tested the output and no extra lines/spaces were created

charts/tfy-llm-gateway/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: tfy-llm-gateway
-version: 0.36.0-rc.1
+version: 0.36.2
```
dunefro, line 3:
```suggestion
version: 0.36.1
```

PR #1304, opened 2025-04-23
Update buildkitd-service sts storage ref

DeeAjayi:


charts/tfy-buildkitd-service/templates/statefulset.yaml
```diff
@@ -148,6 +148,6 @@ spec:
         {{- end }}
         resources:
           requests:
-            storage: {{ .Values.storage.size }}
+            storage: {{- include "buildkitd-service.storageSize" . | indent 1 }}
```
akashg3627, line 151: why do we need indent?

akashg3627 (approved):

PR #1342, opened 2025-05-01
[CI] Publish inframold charts version to 0.1.64

innoavator:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


charts/tfy-k8s-azure-aks-inframold/templates/prometheus-config.yaml
```diff
@@ -30,6 +30,9 @@ spec:
         {{- if .Values.aws }}
           karpenter:
             enabled: {{ .Values.aws.karpenter.enabled }}
+        {{- else }}
+          karpenter:
```
akashg3627, line 34: why would you need a else section?

DeeAjayi: The else section is for other clouds not AWS. This was causing issues with the helm template.

PR #1377, opened 2025-05-15
refactor(ingress.yaml): remove conditional host configuration for ingress rules to simplify template and rely on default behavior

sachincool:


dunefro (approved):
Pleas bump up the Chart.yaml to 0.2.4

PR #1442, opened 2025-06-06
fix(ingress.yaml): remove redundant ssl-redirect action annotation and add ssl-redirect annotation for improved configuration clarity

sachincool:


dunefro (commented):
Please fix the lint - approving to unblock you

dunefro (approved):

PR #1476, opened 2025-06-17
Fix envoy filter apiVersion in crossNamespaceResourcePolicy

gowthamvadisetti:


sachincool (approved):

charts/tfy-configs/files/k8s-manifest-validation-policy.yaml
```diff
@@ -24,7 +24,7 @@ crossNamespaceResourcePolicy:
       - key: apiVersion
         operator: In
         values:
-          - networking.istio.io/v1
+          - networking.istio.io/v1alpha3
```
akashg3627, line 27: we can add multiple here

akashg3627: add both

gowthamvadisetti: updated

PR #1676, opened 2025-07-29
Refactor service account name handling and add image pull secret definition in tfy-llm-gateway template

sachincool:


charts/tfy-buildkitd-service/templates/_helpers.tpl
```diff
@@ -81,10 +81,10 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Create the name of the service account to use
 */}}
 {{- define "buildkitd-service.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create }}
-{{- default (include "buildkitd-service.fullname" .) .Values.serviceAccount.name }}
+{{- if .Values.serviceAccount.name }}
+{{- .Values.serviceAccount.name }}
 {{- else }}
-{{- default "default" .Values.serviceAccount.name }}
+{{- .Values.global.serviceAccount.name }}
```
akashg3627, line 87: do you want to add `-` in end here as well?

charts/tfy-otel-collector/values.yaml
```diff
@@ -10,6 +10,9 @@ global:
   tolerations: []
   ## @param global.nodeSelector [object] Node selector
   nodeSelector: {}
+  ## @param global.serviceAccount.name Service account name
+  serviceAccount:
+    name: "truefoundry"
```
dunefro, line 15: Is this required or will be fetched from parent chart's global

PR #1737, opened 2025-08-11
Update tfy-buildkitd-service chart to version 0.2.11 and add image registry configuration. Modify image references in statefulset and deployment templates to utilize the global image registry setting. Update tfy-clickhouse and tfy-llm-gateway charts to include image registry configurations and adjust image repository paths accordingly.

sachincool:


charts/tfy-clickhouse/values.yaml
```diff
@@ -90,7 +93,7 @@ clickhouse:
   zookeeperPort: '2181'
   labels: {}
   image:
-    repository: tfy.jfrog.io/tfy-mirror/clickhouse/clickhouse-server
+    repository: tfy-mirror/clickhouse/clickhouse-server
```
sachincool, line 96: we control this part, we use values.clickhouse.image.registery ||  .Values.Global.image.registry here now

charts/tfy-clickhouse/values.yaml
```diff
@@ -58,8 +61,8 @@ zookeeper:
   enabled: true
   replicaCount: 1
   image:
-    registry: "tfy.jfrog.io/tfy-mirror"
-    repository: bitnami/zookeeper
+    registry: "tfy.jfrog.io"
```
akashg3627, line 64:
```suggestion
    registry: ""
```

charts/tfy-clickhouse/values.yaml
akashg3627: sure

akashg3627: lets add registry field here?

charts/tfy-clickhouse/values.yaml
sachincool: for this we require it here. 
zookeeper doesnt rely on. global.image.registry

charts/tfy-clickhouse/values.yaml
sachincool: added

PR #1785, opened 2025-08-26
Adding g6f, gr6f, p6e-b200 to gpu nodepool and removing it from other nodepools

dunefro:


chiragjn (approved):

chiragjn (approved):

charts/tfy-karpenter/charts/eks-node-monitoring-agent/values.yaml
```diff
@@ -134,7 +134,63 @@ dcgmAgent:
             # Only schedule on NVIDIA GPU nodes
           - key: node.kubernetes.io/instance-type
```
dunefro, line 135: Shouldn't we replace this to
```
- key: nvidia.com/gpu.deploy.operands
  operator: Exists
```

chiragjn: Can work, but this label is added by us / gpu feature discovery
on bottlerocket we don't add this label (https://github.com/truefoundry/infra-charts/blob/21611124f7cefca516dfadb0b18df9496386c6d8/charts/tfy-karpenter-config/templates/gpu/karpenter-gpu-nodepool.yaml#L10-L24)

We might want monitoring to be independent

dunefro: But adding different instance type is another task to maintain. We can also use `nvidia.com/gpu.count` or `nvidia.com/gpu.present`

chiragjn: > We can also use nvidia.com/gpu.count or nvidia.com/gpu.present

These too will rely on gpu operator components to be working correctly, 

> But adding different instance type is another task to maintain.

I agree but the frequency is like once every 6 months

PR #1835, opened 2025-08-30
Add support for `p5.4xlarge`

chiragjn:


charts/tfy-karpenter/Chart.yaml
```diff
@@ -20,4 +20,4 @@ dependencies:
   name: eks-node-monitoring-agent
   alias: eks-node-monitoring-agent
   repository: file://charts/eks-node-monitoring-agent
-  version: "1.3.0"
+  version: "1.3.1"
```
dunefro, line 23: Let's not do this

charts/tfy-karpenter/charts/eks-node-monitoring-agent/Chart.yaml
```diff
@@ -8,4 +8,4 @@ keywords:
 name: eks-node-monitoring-agent
 sources:
 - https://github.com/aws/eks-node-monitoring-agent
-version: 1.3.0
+version: 1.3.1
```
dunefro, line 11: Same here

charts/tfy-karpenter/Chart.yaml
chiragjn: Got it

PR #2685, opened 2026-04-30
Add sec ctx enable toggle

Ujjwal048:
> [!NOTE]
> **Low Risk**
> Helm templating change with defaults preserving current behavior; risk is limited to users who override these new `enabled` flags or rely on the previous unconditional rendering.
> 
> **Overview**
> Adds **opt-in/opt-out toggles** for emitting `securityContext` in both `tfy-llm-gateway` and `tfy-otel-collector` charts.
> 
> `podSecurityContext`, container `securityContext`, and the custom CA initContainer `securityContext` are now wrapped in `*.enabled` conditionals, and templates omit the `enabled` key when rendering. Values files are updated to include these `enabled` flags (default `true`) and to set `runAsUser`/`runAsGroup` defaults for the custom CA initContainer security context.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2692650ae46f7c51338625951b8d850bbd877bc7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/charts/tfy-llm-gateway/values.yaml
+++ b/charts/tfy-llm-gateway/values.yaml
@@ -120,7 +120,7 @@ nameOverride: ""
 podAnnotations: {}
 ## @param commonAnnotations Common annotations
 commonAnnotations: {}
-## podSecurityContext Pod security context
+## @param podSecurityContext [object] Pod security context
 podSecurityContext:
   enabled: true
   runAsNonRoot: true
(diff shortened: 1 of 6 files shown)


charts/tfy-llm-gateway/values.yaml
```diff
@@ -79,6 +79,9 @@ global:
         sizeLimit: ""
     ## @param global.customCA.securityContext [object] Security context for the custom CA initContainer.
     securityContext:
+      enabled: true
+      runAsUser: 1001
+      runAsGroup: 1001
```
dunefro, line 84: Why have we added runAsGroup now ?

Ujjwal048: Currently it was being inherited from the podsecurity context, if we update that for a specific pod, it might inherit to cutom-ca init as well, so we added it independently now.

dunefro (approved):

PR #3106, opened 2026-08-15
tfy-cloudflared: pin cloudflared egress to Caddy + Cloudflare edge (NetworkPolicy)

shubhamrai1993:
## What

Adds an **egress-only NetworkPolicy** on the cloudflared pods (`cloudflared.networkPolicy.enabled`, default `true`), the outbound counterpart to the Caddy ingress policy from 0.5.0.

cloudflared can now reach **only**:
- **Caddy** (`:80`) — the sole in-cluster target for forwarded tunnel requests
- **DNS** (`:53`)
- **The Cloudflare edge**, pinned by default to Cloudflare's documented Tunnel edge blocks — `198.41.192.0/24`, `198.41.200.0/24`, and their IPv6 `/48`s — on port `7844` TCP/UDP

Nothing else, in-cluster or external. Only `Egress` is restricted, so metrics scraping (ingress) is unaffected. The policy is gated on `caddy.enabled` and is inert on clusters whose CNI doesn't enforce NetworkPolicy.

## Why

Previously a compromised or misconfigured tunnel connector could forward requests to any pod on the cluster network, or open arbitrary outbound connections. This pins the blast radius to Caddy + the Cloudflare edge.

## Config

New values under `cloudflared.networkPolicy`:
- `external.cidrs` — the pinned edge ranges (default). **Empty → falls back** to `0.0.0.0/0` minus `external.fallbackPrivateCIDRs`.
- `external.ports` — default `7844` TCP/UDP.
- `allowDNS`, `allowedEgressTo`, `external.enabled`.

Chart bumped `0.5.0 → 0.6.0`; README regenerated via readme-generator.

## Reviewer notes

- **Hardcoded Cloudflare IPs.** If Cloudflare rotates its edge ranges the tunnel can fail to connect — recover by widening or emptying `external.cidrs`. The enclosing `/24`s (vs. the doc's individual `/32`s) are deliberate, to give headroom.
- **Validated against live devtest cloudflared:** all 8 active QUIC tunnels connect to IPs inside the two pinned `/24`s on port 7844 (colos iad03/08/10/12/16/21) — the allowlist matches real traffic.
- **Enforcement caveat:** the devtest and aws-cle EKS clusters run the VPC CNI with `--enable-network-policy=false` and no Calico/Cilium, so NetworkPolicy is not enforced there — this needs to land on a cluster with enforcement on for the restriction to take effect.
- Relies on the CNI evaluating egress **post-DNAT** (Calico/Cilium do) so the Caddy Service ClusterIP resolves to the Caddy pod the `podSelector` matches — same mechanism the existing Caddy ingress policy already depends on.

---

> [!NOTE]
> **Medium Risk**
> Tightens cluster network egress for the tunnel connector; a wrong or rotated Cloudflare CIDR list can break tunnels on clusters that actually enforce NetworkPolicy.
> 
> **Overview**
> Adds an **egress-only NetworkPolicy** on cloudflared (chart **0.6.0**), the outbound counterpart to Caddy’s ingress policy. In-cluster forwarding is limited to Caddy on port 80; DNS and the Cloudflare edge remain allowed. Ingress (metrics) is unchanged. The policy is rendered only when `caddy.enabled` and `cloudflared.networkPolicy.enabled` (default true).
> 
> External egress is **IP-pinned by default** to Cloudflare’s documented tunnel ranges (`198.41.192.0/24`, `198.41.200.0/24`, IPv6 `/48`s) on port `7844` TCP/UDP. Empty `external.cidrs` falls back to `0.0.0.0/0` minus private CIDRs. Extra rules can be appended via `extraEgress`. If Cloudflare rotates edge IPs, operators must widen or empty the CIDR list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ae9cc8db44dd51234a17adf493214c2ea2f3146. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


charts/tfy-cloudflared/templates/cloudflared/networkpolicy.yaml
```diff
@@ -0,0 +1,71 @@
...
+      {{- with $ext.ports }}
+      ports:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- end }}
+    {{- end }}
+    {{- with .Values.cloudflared.networkPolicy.allowedEgressTo }}
+    - to:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
```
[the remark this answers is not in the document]
shubhamrai1993: Good catch — fixed in d8ff8c0bb.

Replaced the `allowedEgressTo` peer list (which appended a `to`-only rule, reachable on every port) with `extraEgress`: full egress rules (each with its own `to` **and** `ports`) appended verbatim, so the escape hatch stays port-scoped. Default is still `[]`, so the locked-down posture is unchanged.

charts/tfy-cloudflared/templates/cloudflared/networkpolicy.yaml
```diff
@@ -0,0 +1,73 @@
+{{- if and .Values.caddy.enabled .Values.cloudflared.networkPolicy.enabled }}
```
ramantehlan: For cloudflared networkPolicy, we don't need to check if caddy is enabled or not. Practically, caddy will/should be enabled in all cases, but user can still choose to disable it. In any case, we should not depend on caddy to be enabled for this.

charts/tfy-cloudflared/templates/cloudflared/networkpolicy.yaml
```diff
@@ -0,0 +1,73 @@
...
+    {{- include "tfy-cloudflared.commonAnnotations" . | nindent 4 }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "tfy-cloudflared.selectorLabels" . | nindent 6 }}
+  # Egress only: cloudflared may forward tunnel requests to the Caddy router
+  # (and nothing else in-cluster). Ingress is left unrestricted so metrics
+  # scraping keeps working.
+  policyTypes:
+    - Egress
```
ramantehlan:
```suggestion
  policyTypes:
    - Egress
    - Ingress
  ingress:
    - ports:
        - protocol: TCP
           port: {{ .Values.cloudflared.metricsPort }}
       {{- with .Values.cloudflared.networkPolicy.ingress.allowedFrom }}
       from:
           {{- toYaml . | nindent 8 }}
       {{- end }}
```

ramantehlan: Allow from anywhere by default, but only to the metrics port, which serves both metrics and probes.

dunefro (approved):

